PR #38, opened 2020-03-02
Fix Windows CI

duburcqa:
- Improve cmake Python detection
- Fix Python linkage on Windows

It is now possible to build the Python bindings of Jiminy, but not yet for Pinocchio because Eigenpy API has changed.


duburcqa:
@matthieuvigne  enjoy x)

PR #41, opened 2020-03-03
Cleanup telemetry

duburcqa:
- Remove old `EXO_` prefix and hard-coded mesh paths
- Only uses `registerVariable`, `registerConstant` for the telemetry
- Expose `registerConstant` in Python
- Complete exposure of `reset` and `set_state`

Note: Be careful, constants are prefixed by the telemetry object doing the registration, so HighLevelController for the controller.


core/include/jiminy/core/AbstractController.tpp
```diff
@@ -40,11 +40,111 @@ namespace jiminy
...
+            }
             registeredConstants_.emplace_back(fieldName, std::to_string(value));
         }
 
         return returnCode;
     }
+
+    template<>
+    inline result_t AbstractController::registerConstant<char const *>(std::string   const & fieldName,
+                                                                       char  const * const & value)
```
matthieuvigne: Why do you need a special case for ```char const *``` . Can't it be handled  by the templated version as well ?

core/include/jiminy/core/AbstractController.tpp
```diff
@@ -40,11 +40,111 @@ namespace jiminy
...
+    }
+
+    template<>
+    inline result_t AbstractController::registerConstant<char *>(std::string   const & fieldName,
+                                                                 char        * const & value)
+    {
+        return registerConstant<char const *>(fieldName, value);
+    }
+
+    // TODO: Improve to support any type of Eigen object (and const or not)
```
matthieuvigne: Rather "(either const or not)"

core/include/jiminy/core/AbstractController.tpp
duburcqa: I made a new commit that overloads `to_string` instead of using full template specialization. It is both shorter, clearer, and more flexible.

core/include/jiminy/core/AbstractController.tpp
duburcqa: This part of the code has been deleted.

PR #45, opened 2020-03-07
More robust and intuitive API

duburcqa:
- More consistent stepper API with start/step/stop methods.
- More consistent model API to add sensors, motors, and contact points.
- Forbid some undefined behaviours: modifying the model and engine during a simulation, running multiple simulations using the same model using multiple engines.
- Simplify return code logic using early return for methods with errors only at first stack level
- Add flag to specify theoretical or current model initial state. 
- Add a dedicated abstract motor class, and basic class with default friction implementation

Fix https://github.com/Wandercraft/jiminy/issues/46
Fix https://github.com/Wandercraft/jiminy/issues/44
Fix https://github.com/Wandercraft/jiminy/issues/42


python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -1164,20 +1284,28 @@ namespace python
...
                                    (bp::arg("self"), "model", "controller", "callback_handle"))
+
                 .def("reset", &Engine::reset,
-                              (bp::arg("self"), bp::arg("remove_forces")=false))
-                .def("set_state", &Engine::setState,
-                                  (bp::arg("self"), "x_init",
-                                   bp::arg("reset_random_generator")=false,
-                                   bp::arg("remove_forces")=false))
+                              (bp::arg("self"),
+                               bp::arg("reset_random_generator") = false,
```
matthieuvigne: I would set this to `true` by default. 

Rational: calling `reset` then `simulate` twice should give the same result if the random seed in engine options is not changed.

Or this could be removed: such behavior (keeping the random seed) can be done using start/stop in a loop

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -462,6 +462,122 @@ namespace python
...
+                                        bp::return_value_policy<bp::copy_const_reference>()))
+                    .add_property("joint_name", bp::make_function(&AbstractMotorBase::getJointName,
+                                                bp::return_value_policy<bp::copy_const_reference>()))
+                    .add_property("joint_idx", bp::make_function(&AbstractMotorBase::getJointModelIdx,
+                                               bp::return_value_policy<bp::copy_const_reference>()))
+                    .add_property("joint_position_idx", bp::make_function(&AbstractMotorBase::getJointPositionIdx,
+                                                        bp::return_value_policy<bp::copy_const_reference>()))
+                    .add_property("joint_velocity_idx", bp::make_function(&AbstractMotorBase::getJointVelocityIdx,
+                                                        bp::return_value_policy<bp::copy_const_reference>()))
+                    .add_property("torque_limit", bp::make_function(&AbstractMotorBase::getTorqueLimit,
```
matthieuvigne: Add inertia as well, or don't bind any option.

PR #54, opened 2020-03-15
Fix various flaws

duburcqa:
This MR fixes various bugs in the simulation leading to poor computational efficiency because of very small internal step size, and wrong acceleration estimation during the projection on the state space.

- No alterating of the stepper state during the update of the telemetry
- Full precision (double) while parsing the data log file (note that the time accuracy is 1us by design)
- Catch the exception raised by the viewer if Gepetto-viewer is closed during a replay, so that it stops the animation without error.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,7 +2,7 @@
 cmake_minimum_required(VERSION 3.10)
 
 # Set the build version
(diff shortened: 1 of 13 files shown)


duburcqa:
ping @matthieuvigne

PR #55, opened 2020-03-17
Maintainability and consistency improvements

duburcqa:
- Dedicated SOUP folder to improve maintainability
- Create C++ core source subfolders for better segmentation
- Add json import/export utils for configHolder_t
- Refactoring of from/to Python converter using visitor pattern to avoid duplicate code and manual type check
- Rename Model class into Robot and add Model class for better segmentation
- Option getter can not longer raise an error

Note that it does not really break the API apart from minor changes (model -> robot and get_model_options has changed) but it breaks the include paths, so the minor version tag will be increased.

This is related to https://github.com/Wandercraft/jiminy/issues/50 , though it does not fix the issue.


duburcqa:
ping @matthieuvigne (Once again a big MR !)

PR #57, opened 2020-03-20
Add unit tests

matthieuvigne:
Related to https://github.com/Wandercraft/jiminy/issues/39

Provide unit testing in python for Jiminy to verify that Jiminy is capable of simulating simple systems with know analytical answer - and that in that context,Jiminy finds the right answer.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -1,14 +1,12 @@
 name: Linux CI
 
 on:
-  # Trigger the workflow on push or pull request,
-  # but only for the master branch
+  # Trigger the workflow on push on the master branch,
+  # or for any pull request
   push:
     branches:
       - master
   pull_request:
-    branches:
-      - master
 
 jobs:
   build-test-and-deploy:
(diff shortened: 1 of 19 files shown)


matthieuvigne:
@duburcqa This is still WIP, I need to write the pendulum-based tests. I'm opening a PR for you to have a look at the CI, if you are interested: basically I'm trying to launch python code, but it fails to find eigenpy despite the package being installed earlier in the CI. Probably a problem of (virtual) environment in python.

If you don't have time I'll have a look myself, but you know these parts more than I do.

duburcqa:
I will fix the CI, you can focus on the test :)

.github/workflows/linux.yml
```diff
@@ -1,14 +1,6 @@
 name: Linux CI
 
-on:
```
duburcqa: Keep it as is to avoid running the CI for every commit (it is not completely free to run the CI). Or if you prefer, 
```
  push:
    branches:
      - master
  pull_request:
```

.github/workflows/linux.yml
```diff
@@ -93,6 +85,8 @@ jobs:
       run: |
         cd build/unit
         ./unit
+        cd ../../unit_py
+        python3 -m unittest discover
```
duburcqa: Use python instead of python3

duburcqa:
The unit tests are running sucessfully :)

PR #60, opened 2020-03-23
Add Python unit test framework

duburcqa:
Related to https://github.com/Wandercraft/jiminy/issues/39

Provide the framework to write Python unit testing for Jiminy. 

Implement a first unit test verifying that Jiminy is capable of simulating a double spring mass system with known analytical solution - and that in that context, Jiminy finds the right solution.

### DIFF
--- a/.github/workflows/win.yml
+++ b/.github/workflows/win.yml
@@ -127,6 +127,11 @@ jobs:
         path: .\install
     - name: Running unit tests
       run: |
+        $ErrorActionPreference = "Stop"
+        Set-PSDebug -Trace 1
+
+        $RootDir = $Env:GITHUB_WORKSPACE
+
         Set-Location -Path $RootDir\jiminy\build\unit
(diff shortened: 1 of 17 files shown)


.github/workflows/win.yml
```diff
@@ -1,6 +1,12 @@
 name: Windows CI
 
-on: [push, pull_request]
+on:
+  # Trigger the workflow on push on the master branch,
+  # or for any pull request
+  push:
+    branches:
+      - master
+  pull_request:
```
matthieuvigne: If I understand it correctly, the CI will run only after the PR has been accepted ? Unless our original branch is called master, but I never do that (I don't like 'master has been merged onto master') :)

duburcqa: nope, it runs the CI if the destination branch is master! Anyway I removed this condition, so that any PR now triggers the CI.

PR #67, opened 2020-04-01
Add contact constraints support in Jiminy

### LINKED ISSUE #66: Add support for kinematic constaints.
I wish to simulate a segway-like system: basically, a wheel turning on the ground. This cannot be exactly simulated in Jiminy, which only supports punctual contacts with an elastic ground: the 'hack' one could do would be to place many points along the rim of the wheel, and set them all as contact points.

Rather, I want to simulate the contact as an analytical constraint: i.e. the wheel is in perfect contact with the ground and does not slip. This can be done by adding the constraint in the dynamics equation as Lagrange multipliers.

Note that I'm only interested in having a constant contact throughout the simulation: I do not mean to implement 'analytical contact' for a walking robot, that would require an added contact detection framework. What I have in mind is a constant contact throughout the simulation. However, this is already a significant first step in that direction if we ever want to do that.

This is how I see the implementation:

 -  at a 'low-level', contact kinematics is implemented in ```pinocchio``` as ```pinocchio::forwardDynamics(..., J, drift)``` where J is the jacobian of the constraint, and drift the associated drift. This implementation of ```pinocchio::forwardDynamics``` would replace the aba algorithm in ```Engine.computeSystemDynamics```. In practice, ```pinocchio::forwardDynamics``` is likely to be much slower than aba (because it requires computing the inertia matrix, angument it with the jacobian, and do the inverse), thus aba would remain the default option, and Engine would switch to using forwardDynamics only if a contact constraint is being used. 

 - like we did in ```aba```, we will need a custom implementation of ```forwardDynamics``` to support armature inertia. 

 - While I do not need this for my use case, supporting several constraints at once seems a nice feature to have. I think the best way to handle this from the API standpoint is to give a name to each constraint, so adding and removing contact constraints would be like adding and removing sensors or motors.
 - As API, to simulate a contact analytical we need a jacobian, with signature ```matrixN_t jacobian(q)```, and the corresponding drift term with signature ```vectorN_t drif(q, v)```. The main usage would probably be to have a specific frame in contact, however, for a wheel, I want the wheel rim to be in contact, not a fixed point. This requries a 'custom' jacobian to be implemented. So I'm thinking about having two ways to define a constaint: by a simple frame name, or by a callback function.

In the end, here is an overall architecture of what I plan to do:

 - Define signatures for jacobian and drift callback functions:
```c++
typedef matrixN_t (*JacobianFct_t) (model, data, q);
typedef vectorN_t (*DriftFct_t) (model, data, q, v);
```
 - Add class representing the contact:

```c++
class ContactConstraint:
    ContactConstraint(std::string contactFrameName);
    ContactConstraint(JacobianFct_t J, DriftFct_t drift);
```

- Add add/remove methods to robot, like add / remove of motors, sensors, contact frames... work:
    - ```Robot::setContactConstraint(std::string name, ContactConstraint)```

The engine would then switch between both implements of the dynamics depending of the presence of not of ContactContraint. Note that both contact are not mutually exclusive: you could imagine having both intermitent, elastic contacts, and constant constraints (for instance, fix you robot foot on the ground analytically, then make its hand hit a wall).

matthieuvigne:
Fix #66 

Allow the simulation of kinematic contact constraints (i.e. perfect contact) as an additional contact model.


matthieuvigne:
@duburcqa This is not done yet so I'm not expecting a full review, but rather some overall comments about what I've done so far and whether you agree. I've mentioned myself what I think are the main discussion points

core/include/jiminy/core/robot/AbstractConstraint.h
```diff
@@ -0,0 +1,90 @@
...
+        virtual ~AbstractConstraint(void);
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Initialize the constraint on the given model.
+        ///
+        /// \note       This function is called internally when adding a constraint to a robot:
+        ///             there is no need to call it manually.
+        /// \param[in] model    Model on which to apply the constraint.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual hresult_t initialize(Model *model);
```
matthieuvigne, line 49: Initialized is in fact used as an "attach" method. I provide no "detach" method yet but it may be a good thing to have one if only for consistency.

I'm currently exposing ```get_jacobian``` and ```get_drift``` in python for debugging - but the main usage I see is to only create the constraint in python, and add it to a robot - then leave the object be.

core/include/jiminy/core/robot/FixedFrameConstraint.h
```diff
@@ -0,0 +1,90 @@
+///////////////////////////////////////////////////////////////////////////////////////////////
+///
+/// \brief      Class representing a fixed frame constraint.
```
matthieuvigne, line 3: Currently this is the only possible constraint: a fully fixed body. 
This may be generalized by allowing constraint of some axes only (e.g. to simulate an object sliding on a plane).

The other type of constraint I'm interested in is a wheel turning on the ground. At first I though I would do that outside of Jiminy: implement a ```ConstraintFunctor``` class that applies a constraint from a user-specified callback for the jacobian + drift. This allows an arbitrary constraint design from the user.

However, from our discussion I'm thinking ground / wheel is something rather common, so maybe it's a good thing to have an implementation inside Jiminy: this can be done in a generic manner I think, with the user specifying the wheel center, radius and  axis of rotation. Then maybe I don't even need the ```ConstraintFunctor``` class. What do you think ?

core/src/engine/EngineMultiRobot.cc
```diff
@@ -0,0 +1,2257 @@
...
+            matrixN_t J;
+            vectorN_t drift;
+            robot->computeConstraints(q, v, J, drift);
+
+            // TODO: handle external forces.
+            // TODO: rotor inertia.
+
+            // Call forward dynamics.
+            float64_t damping = 1e-12;
+            return pinocchio::forwardDynamics(robot->pncModel_,
```
matthieuvigne, line 2075: Currently the dynamics is limited and does not support two things:

 - external forces (both user-defined and spring contact forces): this can be added to u by the use of the right jacobian as input.
 - rotor inertia: this requires a re-implemtation of forwardDynamics like you did for aba

I'm working on both of these things at the moment.

core/src/robot/Robot.cc
```diff
@@ -1188,6 +1253,30 @@ namespace jiminy
...
+        // Call computeAllTerms to update frame jacobian: this is needed by forwardDynamics anyway.
+        pinocchio::computeAllTerms(pncModel_, pncData_, q, v);
+
+        for (auto & constraint : constraintsHolder_)
+        {
+            matrixN_t J = constraint.second->getJacobian(q);
+            vectorN_t drift = constraint.second->getDrift(q, v);
+            int constraintSize = drift.size();
+            // Concatenate drift and jacobians.
+            jacobianOut.conservativeResize(jacobianOut.rows() + constraintSize, Eigen::NoChange);
```
matthieuvigne, line 1273: This is a point of debate: do we allow the user to change the constraints as the simulation go ? 

Drawback: you don't know in advance the number of constraints, so you have to do dynamic resizing of the matrices, which probably slows down a bit the simulation.

Advantage: even if it's not supported right now, keeping it as is would enable the use of analytical contacts for intermittent constraints: for instance, a walking robot. We "just" need a collision function to determine whether or not the constraint is active at a given time.

core/src/robot/AbstractConstraint.cc
```diff
@@ -0,0 +1,42 @@
...
+namespace jiminy
+{
+    AbstractConstraint::AbstractConstraint() :
+    model_(nullptr),
+    isInitialized_(false)
+    {
+        // Empty on purpose
+    }
+
+    AbstractConstraint::~AbstractConstraint()
```
duburcqa, line 16: use default destructor instead

core/include/jiminy/core/robot/FixedFrameConstraint.h
duburcqa: I think having `ConstraintFunctor` is an interesting feature. 

As you said ground/wheel contact is rather common and perceived as "simple". I don't think and end-user expects having to define it manually. As you said, the input arguments could be "wheel center, radius, axis of rotation" but I think it may be interesting to also be able to set a "width" (equals to zero eventually) so that the wheel can stand on the ground with zero speed without falling.

core/src/robot/Robot.cc
```diff
@@ -1184,6 +1253,30 @@ namespace jiminy
...
+    void Robot::computeConstraints(vectorN_t const & q,
+                                   vectorN_t const & v,
+                                   matrixN_t & jacobianOut,
+                                   vectorN_t & driftOut)
+    {
+        // Give the right number of columns to the jacobian.
+        jacobianOut.resize(0, pncModel_.nv);
+
+        // Call computeAllTerms to update frame jacobian: this is needed by forwardDynamics anyway.
+        pinocchio::computeAllTerms(pncModel_, pncData_, q, v);
```
duburcqa, line 1265: I don't like computeAllTerms. It is too high-level for me and I don't understand what it is actually doing. I would rather prefer to call manually each submethod if possible. Unless it is actually necessary to compute "all" terms.

core/src/robot/Robot.cc
duburcqa: I don't know. Perhaps having a jacobian buffer and checking if the size has changed before doing the conservative resize is the solution. This way, the costly memory allocation will be performed only if necessary, thereby combining the best of both solutions.

core/src/engine/EngineMultiRobot.cc
duburcqa: I guest all the Pinocchio "patches" should be moved in a dedicated helper file.

core/include/jiminy/core/engine/EngineMultiRobot.h
```diff
@@ -691,6 +691,24 @@ namespace jiminy
...
+        ///          or unconstrainted (aba).
+        ///
+        /// \param[in] robot Robot for which to compute the dynamics.
+        /// \param[in] q Joint position.
+        /// \param[in] v Joint velocity.
+        /// \param[in] u Joint torque.
+        /// \param[in] fext External forces applied on the system.
+        /// \return System acceleration.
+        vectorN_t computeAcceleration(std::shared_ptr<Robot> robot,
+                                      vectorN_t const & q,
```
duburcqa, line 707: use Eigen::Ref for q, v.

core/include/jiminy/core/engine/EngineMultiRobot.h
```diff
@@ -691,6 +691,24 @@ namespace jiminy
...
+        ///          with kinematic constraints (using Lagrange multiplier for computing the forces)
+        ///          or unconstrainted (aba).
+        ///
+        /// \param[in] robot Robot for which to compute the dynamics.
+        /// \param[in] q Joint position.
+        /// \param[in] v Joint velocity.
+        /// \param[in] u Joint torque.
+        /// \param[in] fext External forces applied on the system.
+        /// \return System acceleration.
+        vectorN_t computeAcceleration(std::shared_ptr<Robot> robot,
```
duburcqa, line 706: Use `systemDataHolder_t  & system` instead of `std::shared_ptr<Robot>` to have a more consistent API.

core/include/jiminy/core/robot/FixedFrameConstraint.h
```diff
@@ -0,0 +1,90 @@
...
+        /// \brief    Compute and return the jacobian of the constraint.
+        ///
+        /// \note     To avoid duplicate kinematic computation, it is assumed that
+        ///           computeJointJacobians and framesForwardKinematics has already
+        ///           been called on model->pncModel_.
+        ///
+        /// \param[in] q    Current joint position.
+        /// \return         Jacobian of the constraint.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual matrixN_t getJacobian(vectorN_t const & q) const override;
```
duburcqa, line 69: Had final keyword if those methods are not supposed to by override by a child class.

core/include/jiminy/core/robot/Robot.h
```diff
@@ -65,6 +66,38 @@ namespace jiminy
                             Eigen::Ref<vectorN_t const> const & v,
                             Eigen::Ref<vectorN_t const> const & a,
                             vectorN_t                   const & u);
+
+        /// \brief Add a kinematic constraint to the robot.
+        ///
+        /// \param[in] constraintName Unique name identifying the kinematic constraint.
+        /// \param[in] constraint Constraint to add.
+        hresult_t addConstraint(std::string const & constraintName,
```
duburcqa, line 74: I think it would be at Model level instead of Robot.

core/include/jiminy/core/robot/Robot.h
```diff
@@ -65,6 +66,38 @@ namespace jiminy
...
+        /// \param[in] v    Joint velocity.
+        /// \param[out] jacobianOut Output jacobian matrix.
+        /// \param[out] driftOut    Output drift vector.
+        void computeConstraints(vectorN_t const & q,
+                                vectorN_t const & v,
+                                matrixN_t & jacobianOut,
+                                vectorN_t & driftOut);
+
+        /// \brief Returns true if at least one constraint is active on the robot.
+        inline bool hasConstraint()
```
duburcqa, line 96: boot_t

core/include/jiminy/core/robot/Robot.h
```diff
@@ -65,6 +66,38 @@ namespace jiminy
...
+
+        /// \brief Compute jacobian and drift associated to all the constraints.
+        ///
+        /// \note This function internally calls pinocchio::computeAllTerms.
+        ///
+        /// \param[in] q    Joint position.
+        /// \param[in] v    Joint velocity.
+        /// \param[out] jacobianOut Output jacobian matrix.
+        /// \param[out] driftOut    Output drift vector.
+        void computeConstraints(vectorN_t const & q,
```
duburcqa, line 90: Eigen::Ref

core/include/jiminy/core/robot/Robot.h
```diff
@@ -138,6 +171,8 @@ namespace jiminy
 
         std::vector<std::string> motorTorqueFieldnames_;    ///< Fieldnames of the torques of the motors
 
+        std::unordered_map<std::string, std::shared_ptr<AbstractConstraint>> constraintsHolder_;
```
duburcqa, line 174: static_map instead ? It is more efficient. Use only unordered_map if you need to find elements by key repeatedly.

core/src/robot/AbstractConstraint.cc
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    hresult_t AbstractConstraint::refreshProxies()
+    {
+        return hresult_t::SUCCESS;
+    }
+
+    matrixN_t AbstractConstraint::getJacobian(vectorN_t const & q) const
+    {
+        return matrixN_t::Zero(0, model_->pncModel_.nv);
```
duburcqa, line 35: use a internal buffer Jacobian_ instead.

core/src/robot/FixedFrameConstraint.cc
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    hresult_t FixedFrameConstraint::refreshProxies()
+    {
+        return getFrameIdx(model_->pncModel_, frameName_, frameId_);
+    }
+
+    matrixN_t FixedFrameConstraint::getJacobian(vectorN_t const & q) const
+    {
+        matrixN_t J = matrixN_t::Zero(6, model_->pncModel_.nv);
```
duburcqa, line 37: use a internal buffer Jacobian_ instead.

core/src/robot/FixedFrameConstraint.cc
```diff
@@ -0,0 +1,65 @@
...
+        isInitialized_ = returnCode == hresult_t::SUCCESS;
+        return returnCode;
+    }
+
+    hresult_t FixedFrameConstraint::refreshProxies()
+    {
+        return getFrameIdx(model_->pncModel_, frameName_, frameId_);
+    }
+
+    matrixN_t FixedFrameConstraint::getJacobian(vectorN_t const & q) const
```
duburcqa, line 35: As I said, return by const ref

core/src/robot/FixedFrameConstraint.cc
```diff
@@ -0,0 +1,65 @@
...
+    frameId_(0)
+    {
+        // Empty on purpose
+    }
+    FixedFrameConstraint::~FixedFrameConstraint(void)
+    {
+        // Empty on purpose
+    }
+
+    hresult_t FixedFrameConstraint::initialize(Model *model)
```
duburcqa, line 21: spacing

duburcqa: Better use a const * instead if possible (or make it possible if it is not too much trouble)

core/src/robot/FixedFrameConstraint.cc
```diff
@@ -0,0 +1,65 @@
...
+            getFrameJacobian(model_->pncModel_,
+                             model_->pncData_,
+                             frameId_,
+                             pinocchio::LOCAL,
+                             J);
+        }
+        return J;
+    }
+
+    vectorN_t FixedFrameConstraint::getDrift(vectorN_t const & q,
```
duburcqa, line 51: As I said for the jacobian

core/src/robot/FixedFrameConstraint.cc
```diff
@@ -0,0 +1,65 @@
...
+                                             vectorN_t const & v) const
+    {
+        vectorN_t drift = vectorN_t::Zero(6);
+        if (isInitialized_)
+        {
+            drift = getFrameAcceleration(model_->pncModel_,
```
duburcqa, line 57: Here it is even better, since tovector already return a const ref if i'm not mistaken.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
 
+
+    hresult_t Robot::addConstraint(std::string const & constraintName,
+                                   std::shared_ptr<AbstractConstraint> constraint)
+    {
+        hresult_t returnCode = hresult_t::SUCCESS;
+
+        if (constraintsHolder_.count(constraintName) > 0)
+        {
+            std::cout << "Error - Robot::addConstraint - A constraint with name " << constraintName;
```
duburcqa, line 497: with name !?

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+            std::cout << " already exists." << std::endl;
+            returnCode = hresult_t::ERROR_BAD_INPUT;
+        }
+        else
+        {
+            returnCode = constraint->initialize(this);
+            if (returnCode != hresult_t::SUCCESS)
+            {
+                std::cout << "Error - Robot::addConstraint - Fail to initialize constraint. " << std::endl;
```
duburcqa, line 506: Do not print an error message here. constraint->initialize should be doing it.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+                returnCode = hresult_t::ERROR_BAD_INPUT;
+            }
+            else
+            {
+                constraintsHolder_.insert(std::make_pair(constraintName, constraint));
```
duburcqa, line 511: a would prefer `emplace` or (`emplace_back` if available). Never this insert.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+            }
+        }
+        return returnCode;
+    }
+
+    hresult_t Robot::removeConstraint(std::string const & constraintName)
+    {
+        hresult_t returnCode = hresult_t::SUCCESS;
+
+        if (constraintsHolder_.count(constraintName) == 0)
```
duburcqa, line 521: use rather find and save the iterator, in order to pass it to erase.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+            {
+                constraintsHolder_.insert(std::make_pair(constraintName, constraint));
+            }
+        }
+        return returnCode;
+    }
+
+    hresult_t Robot::removeConstraint(std::string const & constraintName)
+    {
+        hresult_t returnCode = hresult_t::SUCCESS;
```
duburcqa, line 519: no need for a returnCode variable here

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+
+        if (constraintsHolder_.count(constraintName) == 0)
+        {
+            std::cout << "Error - Robot::removeConstraint - No constraint with this name exists." << std::endl;
+            returnCode = hresult_t::ERROR_BAD_INPUT;
```
duburcqa, line 524: just early return here.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+        }
+        else
+        {
+            constraintsHolder_.erase(constraintName);
```
duburcqa, line 528: else no longer needed.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,51 @@ namespace jiminy
...
+        }
+        return returnCode;
```
duburcqa, line 530: return success directly.

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -641,6 +642,65 @@ namespace python
...
+        template<class PyClass>
+        class PyConstraintVisit
+        {
+        public:
+            using TConstraint = typename PyClass::wrapped_type;
+
+            static void visit(PyClass& cl)
+            {
+                cl
+                    .def("get_jacobian", &AbstractConstraint::getJacobian)
```
duburcqa, line 664: I would rather use the attribute syntax.

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -641,6 +642,65 @@ namespace python
...
+        ///////////////////////////////////////////////////////////////////////////////
+        /// \brief Expose.
+        ///////////////////////////////////////////////////////////////////////////////
+        static void expose()
+        {
+            bp::class_<AbstractConstraint,
+                       std::shared_ptr<AbstractConstraint>,
+                       boost::noncopyable>("AbstractConstraint", bp::no_init)
+                .def(PyConstraintVisitor());
+            bp::register_ptr_to_python<std::shared_ptr<AbstractConstraint> >(); // Required to handle std::shared_ptr from/to Python (as opposed to boost::shared_ptr)
```
duburcqa, line 691: remove this line

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -641,6 +642,65 @@ namespace python
...
+
+            bp::class_<FixedFrameConstraint, bp::bases<AbstractConstraint>,
+                       std::shared_ptr<FixedFrameConstraint>,
+                       boost::noncopyable>("FixedFrameConstraint", bp::no_init)
+                .def(PyConstraintVisitor())
+                .def("__init__", bp::make_constructor(&PyConstraintVisitor::ConstraintPyFactory,
+                                 bp::default_call_policies(),
+                                 (bp::arg("frame_name"))));
+            bp::register_ptr_to_python<std::shared_ptr<FixedFrameConstraint> >();
```
duburcqa, line 700: remove this line

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -924,6 +984,10 @@ namespace python
                                         bp::arg("sensor_type") = std::string()))
                 .def("get_sensor", &PyRobotVisitor::getSensor,
                                    (bp::arg("self"), "sensor_type", "sensor_name"))
+                .def("add_constraint", &Robot::addConstraint,
+                                     (bp::arg("self"), "name", "constraint"))
+                .def("remove_constraint", &Robot::removeConstraint,
+                                     (bp::arg("self"), "name"))
```
duburcqa, line 990: misalignment

unit_py/test_double_spring_mass.py
```diff
@@ -179,5 +179,122 @@ def external_force(t, q, v, f):
...
+        '''
+        @brief Test kinematic constraint: fixed second mass with a constaint.
+        '''
+        # Set same spings as usual
+        def compute_command(t, q, v, sensor_data, u):
+            u[:] = 0.0
+
+        def internal_dynamics(t, q, v, sensor_data, u):
+            u[:] = - self.k * q - self.nu * v
+        controller = jiminy.ControllerFunctor(compute_command, internal_dynamics)
```
duburcqa, line 192: skip line

unit_py/test_double_spring_mass.py
```diff
@@ -179,5 +179,122 @@ def external_force(t, q, v, f):
...
+        # Add a kinematic constraint on the second mass
+        constraint = jiminy.FixedFrameConstraint("SecondMass")
+        self.robot.add_constraint("fixMass", constraint)
+
+        # Run simulation
+        engine.simulate(self.tf, self.x0)
+
+        log_data, _ = engine.get_log()
+        time = log_data['Global.Time']
+        x_jiminy = np.stack([log_data['HighLevelController.' + s]
```
duburcqa, line 206: In my opinion, '.'.join('HighLevelController', s) is more robust and clear.

unit_py/test_double_spring_mass.py
```diff
@@ -179,5 +179,122 @@ def external_force(t, q, v, f):
...
+        self.robot.add_constraint("world", freeflyer_constraint)
+        fix_mass_constraint = jiminy.FixedFrameConstraint("SecondMass")
+        self.robot.add_constraint("fixMass", fix_mass_constraint)
+
+        # Initialize with zero freeflyer velocity...
+        x_init = np.zeros(17)
+        x_init[7:9] = self.x0[:2]
+        x_init[-2:] = self.x0[2:]
+        # ... and a "random" (but fixed) freeflyer quaternion
+        x_init[:7] = np.array([0.5, -0.8, 42.0, 0.1, 0.5, -0.7, 0.1])
```
duburcqa, line 274: I don't like const random. I would rather reset the random generator of numpy, set the root, and asked it to generate it. Unless your values are not this much random x)

unit_py/test_double_spring_mass.py
```diff
@@ -179,5 +179,122 @@ def external_force(t, q, v, f):
...
+        x_init[3:7] /= np.linalg.norm(x_init[3:7])
+
+        # Run simulation
+        engine.simulate(self.tf, x_init)
+
+        log_data, _ = engine.get_log()
+        time = log_data['Global.Time']
+        x_jiminy = np.stack([log_data['HighLevelController.' + s]
```
duburcqa, line 282: Same as before

core/include/jiminy/core/robot/AbstractConstraint.h
matthieuvigne: The difference for me is that there can be several types of constraints, unlike contact points or external forces. This makes it similar to a motor / sensor. Also, constraint constructors have different parameters (and some have several like the wheel constraint we propose). I don't see an easy way to handle that if the object is created by the Robot class.

But I agree with you that it would make more sense to manage the object by Robot. In a way this is already what is being done using shared_pointers.  The typical use case I see is for the user to create the object in python, then give it to the robot (in fact, I may hide get_jacobian / get_drift from the python bindings such that the only accessible method is the constructor itself). Then if the user releases his pointer, the only remaining instance is in the Robot - such that if you remove the constraint from the robot the object is destroyed. The only time when it's not the case is if the user explicitly kept his instance of shared pointer for whatever reason - and I don't know then if it's justified to destroy the object for him (potentially he wants to keep memory of the constraint he created for attaching it to another robot...).

Of course, the harmful side effect is if you try to attach the same constraints to two robots.

So I propose to have a behavior similar to the sensors with an ```uninitialize``` function (similar to detach) for the constraint. On model removal the constraint is first uninitialized - and you cannot add an already initialized constraint to a robot.

core/include/jiminy/core/robot/AbstractConstraint.h
```diff
@@ -0,0 +1,90 @@
...
+        /// \brief    Compute and return the jacobian of the constraint.
+        ///
+        /// \note     To avoid duplicate kinematic computation, it is assumed that
+        ///           computeJointJacobians and framesForwardKinematics has already
+        ///           been called on model->pncModel_.
+        ///
+        /// \param[in] q    Current joint position.
+        /// \return         Jacobian of the constraint.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual matrixN_t getJacobian(vectorN_t const & q) const;
```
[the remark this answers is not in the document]
matthieuvigne: Currently the object is created by the method. I first I wanted to use const ref, but the thing is the method has to resize the matrix (only itself knows the size of the constraint it wants to apply), and explicitly fills it with zero. And I think that

```
getJacobian(matrixN_t & J)
{
J.resize(x, y);
J.setZero();
} 
```
isn't faster than 
```
getJacobian()
{
J = matrixN_t::Zero(x, y);
return J
}
```

(but I may be wrong, I didn't benchmark.

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -641,6 +642,65 @@ namespace python
...
+        };
+
+    public:
+        template<class PyClass>
+        void visit(PyClass& cl) const
+        {
+            PyConstraintVisit<PyClass>::visit(cl);
+        }
+
+        static std::shared_ptr<FixedFrameConstraint> ConstraintPyFactory(std::string const & frameName)
```
duburcqa, line 677: Do not use a factory.

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
```diff
@@ -641,6 +642,65 @@ namespace python
...
+        }
+
+        ///////////////////////////////////////////////////////////////////////////////
+        /// \brief Expose.
+        ///////////////////////////////////////////////////////////////////////////////
+        static void expose()
+        {
+            bp::class_<AbstractConstraint,
+                       std::shared_ptr<AbstractConstraint>,
+                       boost::noncopyable>("AbstractConstraint", bp::no_init)
```
duburcqa, line 689: use  bp::init<std::string>

core/include/jiminy/core/robot/Robot.h
matthieuvigne: I placed it at Robot level because it currently behaves like a motor / sensor (i.e. an extral object one can attach / detach, not an intrinsic property of the model). So it's linked to the previous debate.

core/src/engine/EngineMultiRobot.cc
```diff
@@ -2093,6 +2091,53 @@ namespace jiminy
...
+        if (robot->hasConstraint())
+        {
+            // Handle kinematic constraints.
+            matrixN_t J;
+            vectorN_t drift;
+            robot->computeConstraints(q, v, J, drift);
+
+            // Project external forces from cartesian space to joint space.
+            vectorN_t uTotal = u;
+            matrixN_t jointJacobian = matrixN_t::Zero(6, robot->pncModel_.nv);
```
duburcqa, line 2109: Same

core/src/engine/EngineMultiRobot.cc
```diff
@@ -2093,6 +2091,53 @@ namespace jiminy
...
+    {
+        if (robot->hasConstraint())
+        {
+            // Handle kinematic constraints.
+            matrixN_t J;
+            vectorN_t drift;
+            robot->computeConstraints(q, v, J, drift);
+
+            // Project external forces from cartesian space to joint space.
+            vectorN_t uTotal = u;
```
duburcqa, line 2108: Do not allocate memory but rather use buffer (for example using a struct ConstraintData_t if they are many)

core/src/engine/EngineMultiRobot.cc
```diff
@@ -2093,6 +2091,53 @@ namespace jiminy
...
+                                            i,
+                                            pinocchio::LOCAL,
+                                            jointJacobian);
+                uTotal += jointJacobian.transpose() * fext[i].toVector();
+            }
+
+            // TODO handle rotor inertia in pinocchio "patch"
+
+            // Call forward dynamics.
+            float64_t damping = 1e-12;
```
duburcqa, line 2124: Define a const on top of the header file.

core/include/jiminy/core/engine/EngineMultiRobot.h
matthieuvigne: I'm not sure to understand the difference at runtime between Eigen::Ref and a standard ref. I though Eigen::Ref only meant to address templating issues (to avoid creating may implementations of the same function), but that it behave like a standard ref. In any case, why should we use Eigen::Ref for q and not for u ? What's the difference between these vectors, they are both vectorN_t const ?

core/include/jiminy/core/robot/AbstractConstraint.h
matthieuvigne: After reading other comments: I'm all the more in favor of handling constraints like motors and sensors, as objects to be plugged in on a robot (not a model), with an initialized / uninitialize mechanism.

duburcqa: > But I agree with you that it would make more sense to manage the object by Robot.

You've convinced me.

I think keeping `get_jacobian / get_drift ` as read-only attribute in Python is nice, at least for debug and quick-and-dirty prototyping.

> Of course, the harmful side effect is if you try to attach the same constraints to two robots.
> So I propose to have a behavior similar to the sensors with an uninitialize function (similar to detach) for the constraint. On model removal the constraint is first uninitialized - and you cannot add an already initialized constraint to a robot.

I don't like the idea of having an `uninitialize` method. I would rather use a mutex (as I did to lock the Robot once the simulation has started). If the mutex is locked, then it is impossible to call `initialize`. Removing the constraint would release the lock, and thereby enable to use `initialize` again.

duburcqa: Just add a get method, to go with add and remove. To be consistent with sensor/motor API.

core/include/jiminy/core/robot/Robot.h
duburcqa: I see your point but even though the mechanism is the same, I don't fully agree that it is more related to the Robot that Model. One could use a Model instead of a Robot if the system does not feature "physical" plug-ins such as motors or sensors. But It should still be possible to add wheel constraint. For example to simulate a simple ball rolling on the ground.

core/include/jiminy/core/robot/Robot.h
matthieuvigne: You mean mpa ? static_map seems to be part of cusom libraries but not in std.

core/src/robot/FixedFrameConstraint.cc
matthieuvigne: const ref doesn't work as is with the python bindings.

```
ue_policy_to_wrap_functions_returning<const Eigen::Matrix<double, -1, 1>&>) (const Eigen::Matrix<double, -1, 1>&)’
     return rc( (tc().*f)(BOOST_PP_ENUM_BINARY_PARAMS_Z(1, N, ac, () BOOST_PP_INTERCEPT)) );
```

unit_py/test_double_spring_mass.py
matthieuvigne: I find direct contatenation clearer than join - especially with '.' as joint word, it's easy to miss it and to misunderstand what the function is doing. Is join really more robust ?

core/include/jiminy/core/robot/FixedFrameConstraint.h
matthieuvigne: I'll implement this in a separate PR so we can merge this already large request.

core/include/jiminy/core/robot/Robot.h
duburcqa: yes, our library :p you can check type.h :)

core/src/robot/FixedFrameConstraint.cc
duburcqa: yes, you have to do something like this:
```
                .add_property("is_initialized", bp::make_function(&Engine::getIsInitialized,
                                                bp::return_value_policy<bp::copy_const_reference>()))
```

core/src/robot/Robot.cc
matthieuvigne: The thing is I don't know the size I will have as input.

In the context of ConstraintFunctor, I don't want the user to have to define a third method (along get_jacobian and get_drift) to give the size of the constraint. I'm also not checking the size of the output, for performance reasons: I expect the user to be reasonable about what he is doing.

Something that could be done would be to check at attach the size of the matrices returned (jacobian should be of size nc x model.vv, drift of size nc, where nc is the constraint dimension). But there can be no guarantee that the user won't modify the size afterward (though that would be unusual I would say). I think we should choose between one of two options (currently I'm doing half of both):

 - trust the user not to change the size of the constraint on the fly. The I can do an initial size computation at attach, removing any runtime checks and allocation. Advantage: fastest solution ; inconvenient: not robust.
 - perform a full check at runtime. Advantage: allow for dynamically changing the size of the constraint by the user. Drawback: a bit slower, because I have to count the elements every time. By handling the memory well I can make it so that allocation is only performed when the size changes, but that would still bring more operations than the first solution.

duburcqa: > perform a full check at runtime. Advantage: allow for dynamically changing the size of the constraint by the user. Drawback: a bit slower, because I have to count the elements every time. By handling the memory well I can make it so that allocation is only performed when the size changes, but that would still bring more operations than the first solution.

Definitely this solution for now. Checking the size should be extremely cheep (and if not we should make sure it is) compare to allocating memory.

unit_py/data/linear_two_masses.urdf
```diff
@@ -7,7 +7,13 @@ This is best described by the following diagram:
 -->
 <?xml version="1.0" ?>
 <robot name="linear_masses">
-    <link name="world"/>
+    <link name="world">
+        <inertial>
+            <origin xyz="0.0 0.0 0.0" rpy="0.0 0.0 0.0"/>
+            <mass value="1.0"/>
```
duburcqa, line 13: Such a heavy world !

unit_py/test_double_spring_mass.py
duburcqa: Yes it is, since the delimiter and telemetry names could easily be exposed in Python, and used to automatically generate the appropriate fieldname, no matter the recursion level.

unit_py/utilities.py
```diff
@@ -0,0 +1,66 @@
...
+    @return A len(time) x dim(x0) array: each line is the solution x at time time[i]
+    '''
+    solver = ode(dynamics)
+    solver.set_initial_value(x0, t = time[0])
+    solver.set_integrator("dopri5")
+    x_sol = [x0]
+    for t in time[1:]:
+        solver.integrate(t)
+        x_sol.append(solver.y)
+    return np.stack(x_sol, axis=0)
```
duburcqa, line 66: skip line

core/include/jiminy/core/robot/AbstractConstraint.h
```diff
@@ -0,0 +1,99 @@
...
+        /// \brief    Compute and return the jacobian of the constraint.
+        ///
+        /// \note     To avoid duplicate kinematic computation, it is assumed that
+        ///           computeJointJacobians and framesForwardKinematics has already
+        ///           been called on model->pncModel_.
+        ///
+        /// \param[in] q    Current joint position.
+        /// \return         Jacobian of the constraint.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual matrixN_t const & getJacobian(vectorN_t const & q);
```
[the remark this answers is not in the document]
matthieuvigne: What do you mean ?

core/include/jiminy/core/engine/AlgorithmsMotorInertia.h
```diff
@@ -0,0 +1,186 @@
...
+#include "pinocchio/algorithm/rnea.hpp"
+#include "pinocchio/algorithm/energy.hpp"
+
+namespace jiminy
+{
+
+    template<typename Scalar, int Options, template<typename, int> class JointCollectionTpl,
+             typename ConfigVectorType, typename TangentVectorType>
+    inline Scalar
+    EngineMultiRobot::kineticEnergy(pinocchio::ModelTpl<Scalar,Options,JointCollectionTpl> const & model,
```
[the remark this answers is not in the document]
matthieuvigne: I don't understand what you ask ? The signature is unchanged, simply moved.

core/include/jiminy/core/engine/AlgorithmsMotorInertia.h
```diff
@@ -0,0 +1,186 @@
+/// Remimplementation of several dynamics algorithms from pinocchio,
+/// adding support for rotor inertia for 1DoF (prismatic, revolue) joints.
+#ifndef JIMINY_ALGORITHMS_MOTOR_INERTIA_H
```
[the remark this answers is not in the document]
matthieuvigne: Yes but they are jiminy's own implementation. The documentation is clear about what is being done ; naming in pinocchio would IMHO imply that we are directly calling pinocchio algorithms - like we do in the rest of the code. On the contrary, we are doing our own, completely independant (from a code point for view) implementation.

core/include/jiminy/core/robot/AbstractConstraint.h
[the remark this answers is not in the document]
matthieuvigne: Ah I see - Eigen::Ref can be used here.

core/include/jiminy/core/robot/AbstractConstraint.h
```diff
@@ -0,0 +1,99 @@
...
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief    Refresh the proxies.
+        ///
+        /// \remark   This method is not intended to be called manually. The Robot to which the
+        ///           motor is added is taking care of it when its own `refresh` method is called.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual hresult_t refreshProxies(void);
+
+        Model const * model_; ///< Model on which the constraint operates.
```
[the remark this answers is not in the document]
matthieuvigne: model_ is already protected.

unit_py/data/linear_two_masses.urdf
matthieuvigne: I know. Gravity is equal to 0.000000000000001g over there - they often neglect it before Coriolis forces. Eastward deviation is a big problem...

matthieuvigne:
ping @duburcqa for review.
I've handled jacobian buffer through the use of two members, I didn't feel this was enough to justify the creation of a true structure, but this can change.

core/include/jiminy/core/engine/AlgorithmsMotorInertia.h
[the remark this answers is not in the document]
duburcqa: Remove `EngineMultiRobot::` and rather use a namespace unless it makes sense to be EngineMultiRobot dependant.

core/include/jiminy/core/engine/AlgorithmsMotorInertia.h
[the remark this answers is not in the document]
duburcqa: Yes but they overload Pinnochio methods, which are completely dependant of it, and using rather complicated implementation because of it. Why not something like PinocchioOverloadAlgorithmsMotorInertia ?

core/include/jiminy/core/robot/Robot.h
```diff
@@ -138,6 +183,10 @@ namespace jiminy
 
         std::vector<std::string> motorTorqueFieldnames_;    ///< Fieldnames of the torques of the motors
 
+        std::vector<std::tuple<std::string, std::shared_ptr<AbstractConstraint>, int>> constraintsHolder_;
```
duburcqa, line 186: Don't use `int`, but rather `int32_t`. and use a struct instead of a tuple in order to have named attribute.

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,75 @@ namespace jiminy
...
+                                             return std::get<0>(element) == constraintName;
+                                         });
+        if (constraintIt == constraintsHolder_.end())
+        {
+            std::cout << "Error - Robot::removeConstraint - No constraint with this name exists." << std::endl;
+            returnCode = hresult_t::ERROR_BAD_INPUT;
+        }
+        else
+        {
+            std::get<1>(*constraintIt)->detach();
```
duburcqa, line 546: As I said above, use struct for clarity.

core/src/robot/Robot.cc
```diff
@@ -1184,6 +1320,34 @@ namespace jiminy
...
+    void Robot::computeConstraints(Eigen::Ref<vectorN_t const> const & q,
+                                   Eigen::Ref<vectorN_t const> const & v)
+    {
+        // Compute joint jacobian.
+        pinocchio::computeJointJacobians(pncModel_, pncData_, q);
+
+        int currentRow = 0;
+        for (auto & constraint : constraintsHolder_)
+        {
+            matrixN_t J = std::get<1>(constraint)->getJacobian(q);
```
duburcqa, line 1332: struct for clarity

duburcqa:
ping @matthieuvigne 

Still some unresolved comments.

core/include/jiminy/core/engine/AlgorithmsMotorInertia.h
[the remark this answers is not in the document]
matthieuvigne: Sure, I get the idea. It's a bit long for a file name though so I simply used PinocchioOverloadAlgorithms

core/include/jiminy/core/robot/AbstractConstraint.h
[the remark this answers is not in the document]
duburcqa: isAttached_ should be private with a const getter (getIsAttached).

core/include/jiminy/core/robot/Robot.h
```diff
@@ -65,6 +66,38 @@ namespace jiminy
...
+        ///
+        /// \param[in] constraintName Unique name identifying the kinematic constraint.
+        /// \param[in] constraint Constraint to add.
+        hresult_t addConstraint(std::string const & constraintName,
+                                std::shared_ptr<AbstractConstraint> constraint);
+
+        /// \brief Remove a kinematic constraint form the system.
+        ///
+        /// \param[in] constraintName Unique name identifying the kinematic constraint.
+        hresult_t removeConstraint(std::string const & constraintName);
```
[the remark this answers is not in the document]
duburcqa: .

core/src/engine/EngineMultiRobot.cc
```diff
@@ -3,14 +3,7 @@
...
 #include "pinocchio/parsers/urdf.hpp"
-#include "pinocchio/algorithm/kinematics.hpp"
-#include "pinocchio/algorithm/frames.hpp"
-#include "pinocchio/multibody/model.hpp"
-#include "pinocchio/multibody/visitor.hpp"
-#include "pinocchio/algorithm/joint-configuration.hpp"
-#include "pinocchio/algorithm/aba.hpp"
-#include "pinocchio/algorithm/rnea.hpp"
-#include "pinocchio/algorithm/energy.hpp"
+#include "pinocchio/algorithm/contact-dynamics.hpp"
```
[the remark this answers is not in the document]
duburcqa: .

core/src/engine/EngineMultiRobot.cc
duburcqa: .

core/src/robot/AbstractConstraint.cc
```diff
@@ -0,0 +1,46 @@
...
+
+#include "jiminy/core/robot/AbstractConstraint.h"
+
+
+namespace jiminy
+{
+    AbstractConstraint::AbstractConstraint() :
+    model_(nullptr),
+    isAttached_(false),
+    jacobian_(matrixN_t::Zero(0,0)),
```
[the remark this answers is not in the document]
duburcqa: .

core/src/robot/AbstractConstraint.cc
```diff
@@ -0,0 +1,46 @@
...
+    drift_(vectorN_t::Zero(0))
```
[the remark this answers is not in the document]
duburcqa: .

core/src/robot/AbstractConstraint.cc
```diff
@@ -0,0 +1,46 @@
...
+        return hresult_t::SUCCESS;
+    }
+
+    void AbstractConstraint::detach()
+    {
+        model_ = nullptr;
+        isAttached_ = false;
+    }
+
+    hresult_t AbstractConstraint::refreshProxies()
```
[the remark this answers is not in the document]
duburcqa: .

core/src/robot/Robot.cc
```diff
@@ -474,6 +486,66 @@ namespace jiminy
...
+        auto constraintIt = std::find_if(constraintsHolder_.begin(),
+                                         constraintsHolder_.end(),
+                                         [&constraintName](auto const & element)
+                                         {
+                                             return element.first == constraintName;
+                                         });
+        if (constraintIt != constraintsHolder_.end())
+        {
+            std::cout << "Error - Robot::addConstraint - A constraint with name " << constraintName;
+            std::cout << " already exists." << std::endl;
```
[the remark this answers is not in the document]
duburcqa: .

core/src/robot/Robot.cc
duburcqa: .

core/src/robot/Robot.cc
duburcqa: .

core/src/robot/Robot.cc
duburcqa: .

core/src/robot/Robot.cc
duburcqa: .

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
duburcqa: .

python/jiminy_pywrap/include/jiminy/python/Jiminy.h
duburcqa: .

core/src/robot/Robot.cc
matthieuvigne: I'm doing this for symmetry with other methods like detachMotor or detachSensors which work with the same principle of having only one return at the end.

core/include/jiminy/core/robot/AbstractConstraint.h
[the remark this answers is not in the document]
matthieuvigne: I don't really see why isAttached_ in particular should be private - I need access to this variable in a child class, a getter and setter for a simple bool seems kind of overkill.

core/include/jiminy/core/robot/Robot.h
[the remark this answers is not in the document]
matthieuvigne: I know, but is this really something we need ? The constraint is kind of passive to me: it's there, but there isn't much things one would want to do with it beside creating it.

core/src/engine/EngineMultiRobot.cc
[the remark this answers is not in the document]
matthieuvigne: pinocchio::forwardDynamics with kinematic constraints comes from this file.

core/src/robot/AbstractConstraint.cc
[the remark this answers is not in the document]
matthieuvigne: Because binding a pure virtual function requires extra work so as not to get an undefined symbol at library loading. I took the lazy option to make a concrete implementation instead.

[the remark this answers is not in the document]
duburcqa: Sure, but this method should not have binding since it is only used internally. It is only public to avoid complex interface architecture.

core/include/jiminy/core/robot/Robot.h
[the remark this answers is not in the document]
duburcqa: It could be useful for debug a step-by-step simulation, when the contraint is not accessible because it has been created by another method (as we are doing for atalante). Getting access to the contraint is necessary to extract the Jacobian,.

core/include/jiminy/core/robot/AbstractConstraint.h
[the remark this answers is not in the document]
duburcqa: I though the derived class only needed a read-only access to this member. It would be the case to force the user to call the base implementation setting the flag, thereby ensuring the desired minimal attaching behaving.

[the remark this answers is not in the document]
duburcqa: Of course no setter should be implemented, otherwise it is irrelevant not to give access to the member directly.

core/src/robot/AbstractConstraint.cc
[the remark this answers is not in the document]
duburcqa: Only methodq useful for the user should be binded in Python.

core/include/jiminy/core/robot/AbstractConstraint.h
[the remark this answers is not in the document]
matthieuvigne: No the derived needs access, because initialize or refreshProxies may fail (e.g. if the frame name does not exist in the model) - this is something the parent has no knowledge about.

core/include/jiminy/core/robot/Robot.h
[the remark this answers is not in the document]
matthieuvigne: True, adding function.

core/src/robot/AbstractConstraint.cc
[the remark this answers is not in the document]
matthieuvigne: Even without binding the method, the simple fact that the class is binded poses a problem - because it means that the user can instanciate the class, but the method itself is undefined.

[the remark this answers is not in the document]
matthieuvigne: Oh ok I get it - I though not having a definition would make the method pure virtual, but no you need to explicitly at ```= 0``` to the definition.  And you need to prvent python initialization with bp::no_init

PR #82, opened 2020-04-15
[python] Add generic plotting tool for jiminy logs.

matthieuvigne:
To make jiminy even nicer to use, this PR proposes a simple python tool, `jiminy_plot`, to plot the content of a logfile rapidly using matplotlib. Below is an example of rendering (actual data is meaningless):

```
jiminy_plot  /tmp/example_sagittal.data H*.c*Pos*Body:H*.c*Vel*Body H*.energy
```

example

### DIFF
--- a/python/jiminy_py/src/jiminy_py/__init__.py
+++ b/python/jiminy_py/src/jiminy_py/__init__.py
@@ -15,3 +15,5 @@ except ImportError:
 
 # Import core submodule
 from . import core
+
+from .utils import read_log
--- a/python/jiminy_py/src/jiminy_py/__init__.py
+++ b/python/jiminy_py/src/jiminy_py/__init__.py
@@ -15,5 +15,3 @@ except ImportError:
 
 # Import core submodule
 from . import core
-
-from .utils import read_log
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -10,11 +10,10 @@ from .state import State
 from .core import Engine, Robot
 
 def extract_state_from_simulation_log(log_data:tp.Dict, robot:Robot):
-    """
-    @brief      Extract a trajectory object using from raw simulation data.
+    """ Extract a trajectory object using from raw simulation data.
 
     @details    Extract time, joint positions and velocities evolution from log.
-    .
+
     @remark     Note that the quaternion angular velocity vectors are expressed
                 it body frame rather than world frame.
 
@@ -44,8 +43,7 @@ def extract_state_from_simulation_log(log_data:tp.Dict, robot:Robot):
 
 
 def read_log(filename:str) -> tp.Tuple[tp.Dict, tp.Dict]:
-    """
-    Read a logfile from jiminy. This function supports both text (csv)
+    """ Read a logfile from jiminy. This function supports both text (csv)
     and binary log.
 
     Parameters:
@@ -82,8 +80,7 @@ def read_log(filename:str) -> tp.Tuple[tp.Dict, tp.Dict]:
     return data_dict, constants_dict
 
 def is_log_binary(filename):
-    """
-    From https://stackoverflow.com/a/11301631/4820605.
+    """ From https://stackoverflow.com/a/11301631/4820605.
     Return true if the given filename appears to be binary.
     File is considered to be binary if it contains a NULL byte.
     """
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -3,7 +3,7 @@
 ## @file jiminy_py/log.py
 
 import numpy as np
-from pandas import read_csv
+from csv import DictReader
 import typing as tp
 
 from .state import State
@@ -54,7 +54,7 @@ def read_log(filename:str) -> tp.Tuple[tp.Dict, tp.Dict]:
         - A dictionnary containing the logged values, and a dictionnary
         containing the constants.
     """
-    if is_binary(filename):
+    if is_log_binary(filename):
         # Read binary file using C++ parser.
         data_dict, constants_dict = Engine.read_log_binary(filename)
     else:
@@ -66,19 +66,26 @@ def read_log(filename:str) -> tp.Tuple[tp.Dict, tp.Dict]:
                 c_split = c.split('=')
                 # Remove line end for last constant.
                 constants_dict[c_split[0]] = c_split[1].strip('\n')
-        # Read data from the log file, skipping the first line (the constants).
-        data = read_csv(filename, skiprows=1).to_dict(orient='list')
+            # Read data from the log file, skipping the first line (the constants).
+            data = {}
+            reader = DictReader(log)
+            for key, value in reader.__next__().items():
+                data[key] = [value]
+            for row in reader:
+                for key, value in row.items():
+                    data[key].append(value)
+            for key, value in data.items():
+                data[key] = np.array(value, dtype=np.float64)
         # Convert every element to array to provide same API as the C++ parser,
         # removing spaces present before the keys.
         data_dict = {k.strip() : np.array(v) for k,v in data.items()}
     return data_dict, constants_dict
 
-def is_binary(filename):
+def is_log_binary(filename):
     """
     From https://stackoverflow.com/a/11301631/4820605.
     Return true if the given filename appears to be binary.
     File is considered to be binary if it contains a NULL byte.
-    FIXME: This approach incorrectly reports UTF-16 as binary.
     """
     with open(filename, 'rb') as f:
         for block in f:
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -10,10 +10,11 @@ from .state import State
 from .core import Engine, Robot
 
 def extract_state_from_simulation_log(log_data:tp.Dict, robot:Robot):
-    """ Extract a trajectory object using from raw simulation data.
+    """
+    @brief      Extract a trajectory object using from raw simulation data.
 
     @details    Extract time, joint positions and velocities evolution from log.
-
+    .
     @remark     Note that the quaternion angular velocity vectors are expressed
                 it body frame rather than world frame.
 
@@ -43,7 +44,8 @@ def extract_state_from_simulation_log(log_data:tp.Dict, robot:Robot):
 
 
 def read_log(filename:str) -> tp.Tuple[tp.Dict, tp.Dict]:
-    """ Read a logfile from jiminy. This function supports both text (csv)
+    """
+    Read a logfile from jiminy. This function supports both text (csv)
     and binary log.
 
     Parameters:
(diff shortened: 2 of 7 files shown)


duburcqa:
Amazing !

python/jiminy_py/src/jiminy_py/plotter.py
```diff
@@ -0,0 +1,74 @@
...
+                      "Simply specify a list of fields to plot, separated by a colon for plotting on the same subplot.\n" +\
+                      "Example: h1 h2:h3 generates two subplots, one with h1, one with h2 and h3.\n" + \
+                      "Regular expressions can be used. Enter no plot command (only the file name) to view the list of fields available inside the file."
+
+    parser = argparse.ArgumentParser(description = description_str, formatter_class = argparse.RawTextHelpFormatter)
+    parser.add_argument("input", help = "Input logfile.")
+    main_arguments, plotting_commands = parser.parse_known_args()
+
+    # Load log file.
+    log_data, _ = Engine.read_log(main_arguments.input)
```
duburcqa, line 21: either we should be able to deal with the csv file, or completely remove the support.

python/jiminy_py/src/jiminy_py/plotter.py
```diff
@@ -0,0 +1,74 @@
...
+    t = log_data['Global.Time']
+    # Plot each element.
+    for i in range(n_plot):
+        for name in plotted_elements[i]:
+            axs[i].plot(t, log_data[name], label = name)
+    # Add legend to upper left corner.
+    for ax in axs:
+        ax.legend(bbox_to_anchor=(1.0, 1.0), loc = 1)
+        ax.grid()
+    plt.subplots_adjust(bottom=0.05, top=0.98, left=0.06, right=0.98, wspace=0.1, hspace=0.05)
```
duburcqa, line 69: Do you know the automatic `tight_layout` ?
https://matplotlib.org/3.2.1/tutorials/intermediate/tight_layout_guide.html

duburcqa:
Perhaps it would be great to label x-axis, what do you think ?

duburcqa:
PS: Don't mind the CI failing

python/jiminy_py/src/jiminy_py/plotter.py
matthieuvigne: This is an issue that was present before this PR: only binary  log can be read.

I think  csv is a familiar format for many people so it's nice to have it there. But clearly we both use the binary log as the main format.

A solution if we don't want to fully support two formats at the same time (which I think is a good idea) is to propose CSV as an export option only - make it clear in the API that there is a 'writeLog/readLog' pair for Jiminy's binary format, and an 'exportToCSV' tool for people who need it - that would be for interfacing with other tools only, standard jiminy project would use the binary file. What do you think ?

duburcqa: I didn't mean to add support in C++, but at least in Python. We should ad an helper to read log in `log.py`. It would be only 2 lines or code using the native method `csv.DictReader` https://docs.python.org/3/library/csv.html#csv.DictReader.

Then, it would be great to have an optional parameter `is_binary` to your method (as for `write_log`) switching because the 2 reading methods.

duburcqa: I agree it was present before your PR, but I think it does not require a lot of work but may greatly improve the support the csv format.

matthieuvigne: Agreed, doing it in python is really fast so it makes sense. I was thinking about doing it in C++ at first because `Engine.read_log` is in C++.

How about I write a ```jiminy_py.read_log(logfile)``` function in python, that detects if the log is csv or binary (shouldn't be hard to do it automatically), and calls the right function accordingly ? I would rename the orginal, C++-binded method to ```Engine.read_log_binary``` to avoid confusion.

duburcqa: It would be amazing :)  (or perhaps `jiminy_py.log.read_log` but whatever)

duburcqa: Have a look at this https://stackoverflow.com/a/11301631/4820605

python/jiminy_py/src/jiminy_py/plotter.py
matthieuvigne: I'm often disapointed by what tight_layout does. I don't know if I'm using maplotlib badly, but here is the output I get:

example_sagittal data

I don't know what is causing this big space in the middle...
I'm not fond of hard-coding setting either ; however I've tested this software, with a few (say less than 12) cuves the spacing works well. Above that, well the plot will be ugly anyway...

matthieuvigne:
> Perhaps it would be great to label x-axis, what do you think ?

I don't know, since we have several plots side by side you have to repeat the legend and I don't like the look of that. 
This is for quick plotting anyway, I expect users to make their custom scripts if they want really nice looking plots.  As is, I would rather have something lighweight, leaving as much room as possible for the plots rather than the legends.

python/jiminy_py/src/jiminy_py/plotter.py
duburcqa: hum strange.. usually the result is pretty good for me but I don't remember if I combine it with something else. Anyway it was just for information. What you have done look nice as is 👍

duburcqa:
> I don't know, since we have several plots side by side you have to repeat the legend and I don't like the look of that.

I see your point and I'm ok with it.

python/jiminy_py/src/jiminy_py/plotter.py
matthieuvigne: I did things a bit differently based on other examples: I try to read the log as text and if a Unicode parsing exception is raised, I switch to binary. But I can implement your solution if you prefer.

python/jiminy_py/src/jiminy_py/__init__.py
```diff
@@ -15,3 +15,5 @@
 
 # Import core submodule
 from . import core
+
+from .utils import read_log
```
duburcqa, line 19: Why not implement `read_log` in `log.py` ?

python/jiminy_py/src/jiminy_py/plotter.py
```diff
@@ -0,0 +1,75 @@
+# Plot data from a jiminy log file. See inline help for more info.
+
+import argparse
+import numpy as np
+import matplotlib.pyplot as plt
+from fnmatch import filter
+
+import jiminy_py
```
duburcqa, line 8: strange to import itself

python/jiminy_py/src/jiminy_py/plotter.py
```diff
@@ -0,0 +1,75 @@
...
+                      "Simply specify a list of fields to plot, separated by a colon for plotting on the same subplot.\n" +\
+                      "Example: h1 h2:h3 generates two subplots, one with h1, one with h2 and h3.\n" + \
+                      "Regular expressions can be used. Enter no plot command (only the file name) to view the list of fields available inside the file."
+
+    parser = argparse.ArgumentParser(description = description_str, formatter_class = argparse.RawTextHelpFormatter)
+    parser.add_argument("input", help = "Input logfile.")
+    main_arguments, plotting_commands = parser.parse_known_args()
+
+    # Load log file.
+    log_data, _ = jiminy_py.read_log(main_arguments.input)
```
duburcqa, line 21: why not in `log` submodule ?

python/jiminy_py/src/jiminy_py/__init__.py
duburcqa: By the way, `utils.py` is missing.

python/jiminy_py/src/jiminy_py/plotter.py
duburcqa: I have to admit I'm not a huge fun of `try ... catch ...` mechanism when a more intelligible criterion can be used. Moreover here the criterion is robust and easy to implement.

duburcqa:
> PS: Don't mind the CI failing

Mind it now :p

python/jiminy_py/src/jiminy_py/__init__.py
matthieuvigne: I just didn't see log.py :)

python/jiminy_py/src/jiminy_py/plotter.py
matthieuvigne: From the comment in the code ('    FIXME: This approach incorrectly reports UTF-16 as binary.') it doesn't look super robust, but the try catch clearly wasn't either. I've changed to the proposed solution.

duburcqa: Yes it is not, in a sense that it cannot be used in the general case, but we now what we expect to be a binary log file in our case, and what this method is checking will never change in the future. That's why I said it would robust :)

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -39,3 +41,47 @@ def extract_state_from_simulation_log(log_data, robot):
...
+                # Remove line end for last constant.
+                constants_dict[c_split[0]] = c_split[1].strip('\n')
+        # Read data from the log file, skipping the first line (the constants).
+        data = read_csv(filename, skiprows=1).to_dict(orient='list')
+        # Convert every element to array to provide same API as the C++ parser,
+        # removing spaces present before the keys.
+        data_dict = {k.strip() : np.array(v) for k,v in data.items()}
+    return data_dict, constants_dict
+
+def is_binary(filename):
```
duburcqa, line 76: `is_binrary_log` would be better, since it is the only recommended use case of this method (which lack of robustness otherwise)

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -39,3 +41,47 @@ def extract_state_from_simulation_log(log_data, robot):
...
+    """
+    From https://stackoverflow.com/a/11301631/4820605.
+    Return true if the given filename appears to be binary.
+    File is considered to be binary if it contains a NULL byte.
+    FIXME: This approach incorrectly reports UTF-16 as binary.
```
duburcqa, line 81: Remove this comment since it is not relevant in our case.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -3,11 +3,13 @@
 ## @file jiminy_py/log.py
 
 import numpy as np
+from pandas import read_csv
```
duburcqa, line 6: I don't like the idea to add pandas as a dependency for only this feature, which is able on the standard library `csv`.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -39,3 +41,47 @@ def extract_state_from_simulation_log(log_data, robot):
...
+        # Read text csv file.
+        constants_dict = {}
+        with open(filename, 'r') as log:
+            consts = next(log).split(', ')
+            for c in consts:
+                c_split = c.split('=')
+                # Remove line end for last constant.
+                constants_dict[c_split[0]] = c_split[1].strip('\n')
+        # Read data from the log file, skipping the first line (the constants).
+        data = read_csv(filename, skiprows=1).to_dict(orient='list')
```
duburcqa, line 70: rather use `data = reader = csv.DictReader(log)` (to put inside the `with` statement)

matthieuvigne:
@duburcqa It's nice to have code quality review for python !
However I don't understant what codacy is asking right now. I though it wanted me to start the comment on the line the """ characters, but that also throws an error saying that comment should start on the second line only...

matthieuvigne:
Ok I think there is something misconfigured in Codacy CI: I ran into the error mentionned in  https://stackoverflow.com/questions/45990301/pep257-d212-and-d213-conflicts - I think we have both D212 and 213 activated, hence the constant conflict (now I have a D212 warning but I tried moving everything to the first line in the previous commit and got a D213 error).

duburcqa:
Codacy is not required for now and I agree some warnings are not very clear :/ I'm OK to merge as is if you want !

PR #88, opened 2020-04-20
Various improvements

duburcqa:
- [jiminy core] Add jointType attribute to the motor. Only support linear and rotary joints for motors.
- [jiminy core] Replace Torque by Effort to be appropriate for both rotary and linear joints.
- [jiminy core] Add prefix 'joints' to the ambiguous fields (Position|Velocity)Limit*.
- [jiminy core] Define more appropriate position/velocity/effort limits for non-physical joints.

____

- [jiminy_py C++] Improve pinocchio/eigenpy 'submodule' management.
- [jiminy_py C++] Add getters and expose sensor type specific properties. Fix wrong exposure of the sensors names.
- [jiminy_py C++] Fix Robot class used instead of Model in Jiminy bindings.

____

- [jiminy_py Python] Improve robustness of viewer to handle closed gepetto-gui during rendering.
- [jiminy_py Python] Enable to render the initial state of EngineAsynchronous.

____

- [gym_jiminy] Refactoring of Gym Jiminy to extract more information from the model instead of using hardcoded values.
-  [gym_jiminy] The observation space now corresponds to the sensors and state spaces by default.
-  [gym_jiminy] Use Global variable to define "universal bounds" to overwrite jiminy inf bounds if desired. 

____

- [jiminy_plot] Move plotter.py in log.py since they are related. Remove Python3 only typing.
- [jiminy_plot] Prepend expressions with ':' enable to plot every matched fields in the same subplot.

For example
```
jiminy_plot /tmp/traj.data :*.FX:*.FY:*.FZ :*.Q :*.V *energy
```

image
____

- [doc] Update readme.

### DIFF
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -104,7 +104,7 @@ def plot_log():
         "Example: h1 h2:h3:h4 generates two subplots, one with h1, one with h2, h3, and h4.\n" + \
         "Wildcard token '*' can be used. In such a case:\n" + \
         "- If *h2* matches several fields : each field will be plotted individually in subplots. \n" + \
-        "- If :*h2* matches several fields : each field will be plotted in the same subplot. \n" + \
+        "- If :*h2* or :*h2*:*h3*:*h4* matches several fields : each field will be plotted in the same subplot. \n" + \
         "- If *h2*:*h3*:*h4* matches several fields : each match of h2, h3, and h4 will be plotted jointly in subplots.\n" + \
         "  Note that if the number of matches for h2, h3, h4 differs, only the minimum number will be plotted.\n" + \
         "\nEnter no plot command (only the file name) to view the list of fields available inside the file."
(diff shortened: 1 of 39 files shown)


duburcqa:
ping @matthieuvigne

matthieuvigne:
Do you really want to support python2.7 now that it's officially deprecated ?  Removing typing hints because a deprecated version of python doesn't support it seems kind of weird...

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -81,14 +97,79 @@ def read_log(filename:str) -> tp.Tuple[tp.Dict, tp.Dict]:
...
+    if len(plotting_commands) == 0:
+        print("Available data:")
+        print("\n - ".join(log_data.keys()))
+        exit(0)
+
+    # Parse plotting arguments.
+    plotted_elements = []
+    for cmd in plotting_commands:
+        # Check that the command is valid, i.e. that all elements exits. If it is the case, add it to the list.
+        same_subplot = (cmd[0] == ':')
```
matthieuvigne, line 130: Nice feature :)

duburcqa:
> Do you really want to support python2.7 now that it's officially deprecated ? Removing typing hints because a deprecated version of python doesn't support it seems kind of weird...

I don't know. But we need to decide if we really want to completely drop the support officially or not, and if so, update the Readme. According to the statistics, between 5% and 15% of the researchers are still using Python2. Personally, I don't think typing is enough to drop support before the end of 2020. What do you think ?

matthieuvigne:
No I guess not - the best would be to have a separate python2 and python3 version, but clearly we don't want to maintain that. So let's stick to python2 for now...

PR #98, opened 2020-04-30
[core] Fix bug in FixedFrameConstraint resetProxies.

matthieuvigne:
resetProxies was not updating the model size, leading to issues when
switching from rigid to flexible model.

### DIFF
--- a/core/src/robot/FixedFrameConstraint.cc
+++ b/core/src/robot/FixedFrameConstraint.cc
@@ -68,6 +68,8 @@ namespace jiminy
 
     hresult_t FixedFrameConstraint::refreshProxies()
     {
+        // Resize the jacobian to the model dimension.
+        jacobian_.resize(6, model_->pncModel_.nv);
         return getFrameIdx(model_->pncModel_, frameName_, frameIdx_);
     }
 }


duburcqa:
You should increase the release tag. If you wish, you could gather your PRs to avoid increasing it multiple times for small modifications.

matthieuvigne:
For me there is no need to make a new tag for every PR ; we just make a new release tag when there is a sufficient  amount of content. Otherwise we will loose a lot of time constantly increasing tags on PRs.

Possibly we should work with a dev branch if you want master to always point to a tagged version, like what is done e.g. in pinocchio.

duburcqa:
If you want. But in this case we NEED a dev branch, since we have to increase the version tag for each commit on the master branch otherwise the CI will fail (because pypi deploy does not work if the version tag already exists)

matthieuvigne:
Ah ok, I didn't know we had this constraint. 

Personally I prefer making independent PRs for independant features (huge 'various improvements' PRs are difficult to follow), but I can understand if you don't want to handle a dev branch (in particular for these 1-line PRs). I'll merge them all then and that will be good for now.

matthieuvigne:
It would be nice if we could have a single version number for the whole project so we don't have to change it in 5 places, but I don't know if it's easily doable by both cmake / setuptools at the same time.

matthieuvigne:
Closing as merged in https://github.com/Wandercraft/jiminy/pull/93

PR #111, opened 2020-05-16
[jiminy] Allow change in the unit used to log the simulation.

matthieuvigne:
Add `Global.TIME_UNIT` constant to specify the unit in which time is logged, this enables the user to change the precision with which logging is done.

### DIFF
--- a/core/include/jiminy/core/telemetry/TelemetryData.h
+++ b/core/include/jiminy/core/telemetry/TelemetryData.h
@@ -20,12 +20,11 @@ namespace jiminy
     std::string const NUM_INTS("NumIntEntries=");        ///< Number of integers in the data section.
     std::string const NUM_FLOATS("NumFloatEntries=");    ///< Number of floats in the data section.
     std::string const GLOBAL_TIME("Global.Time");        ///< Special column
+    std::string const TIME_UNIT("Global.TIME_UNIT");          ///< Special constant
     std::string const START_CONSTANTS("StartConstants"); ///< Marker of the beginning the constants section.
     std::string const START_COLUMNS("StartColumns");     ///< Marker of the beginning the columns section.
     std::string const START_LINE_TOKEN("StartLine");     ///< Marker of the beginning of a line of data.
     std::string const START_DATA("StartData");           ///< Marker of the beginning of the data section.
-    std::string const START_INPUT_FRAME("INPUT_START");  ///< Marker of the beginning of a input ethercat frame.
-    std::string const START_OUTPUT_FRAME("OUTPUT_START");///< Marker of the beginning of a output ethercat frame.
 
     std::size_t const CONSTANTS_MEM_SIZE = 16U * 1024U;
     std::size_t const INTEGERS_MEM_SIZE  = 32U * 1024U;
--- a/core/include/jiminy/core/telemetry/TelemetryRecorder.h
+++ b/core/include/jiminy/core/telemetry/TelemetryRecorder.h
@@ -40,7 +40,7 @@ namespace jiminy
 
         /// \brief Get the maximum time that can be logged with the current precision.
         /// \return Max time, in second.
-        float64_t getMaximumLogTime() const;
+        float64_t getMaximumLogTime(void) const;
 
         ////////////////////////////////////////////////////////////////////////
         /// \brief Reset the recorder.
--- a/core/include/jiminy/core/telemetry/TelemetryRecorder.h
+++ b/core/include/jiminy/core/telemetry/TelemetryRecorder.h
@@ -38,6 +38,10 @@ namespace jiminy
 
         bool_t const & getIsInitialized(void);
 
+        /// \brief Get the maximum time that can be logged with the current precision.
+        /// \return Max time, in second.
+        float64_t getMaximumLogTime() const;
+
         ////////////////////////////////////////////////////////////////////////
         /// \brief Reset the recorder.
         ////////////////////////////////////////////////////////////////////////
--- a/core/include/jiminy/core/telemetry/TelemetryRecorder.h
+++ b/core/include/jiminy/core/telemetry/TelemetryRecorder.h
@@ -29,8 +29,12 @@ namespace jiminy
 
         ////////////////////////////////////////////////////////////////////////
         /// \brief Initialize the recorder.
+        /// \param[in] telmetryData Data to log.
+        /// \param[in] timeUnit Unit with which the time will be logged
+        ///                     (note that time is logged as an int).
         ////////////////////////////////////////////////////////////////////////
-        hresult_t initialize(TelemetryData const * telemetryData);
+        hresult_t initialize(TelemetryData * telemetryData,
+                             float64_t const& timeLoggingPrecision);
 
         bool_t const & getIsInitialized(void);
 
@@ -89,6 +93,7 @@ namespace jiminy
 
         char_t const * floatsAddress_;      ///< Address of the float data section.
         int64_t floatSectionSize_;          ///< Size in byte of the float data section.
+        float64_t timeLoggingPrecision_;    ///< Precision to use when logging the time.
     };
 }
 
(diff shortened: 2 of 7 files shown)


core/src/engine/EngineMultiRobot.cc
```diff
@@ -1115,9 +1115,8 @@ namespace jiminy
                     while (tNext - t > EPS)
                     {
                         /* Adjust stepsize to end up exactly at the next breakpoint,
-                           prevent steps larger than dtMax, and make sure that dt is
-                           multiple of TELEMETRY_TIME_DISCRETIZATION_FACTOR whenever
-                           it is possible, to reduce rounding errors of logged data. */
+                           prevent steps larger than dtMax, trying to reach multiples of
+                           STEPPER_MIN_TIMESTEP whenever possible. */
```
matthieuvigne, line 1119: @duburcqa this deserves some thoughts.
You implemented a mechanism to synchronized simulation timestep with the log unit. Now the log unit can change. I can keep the mechanism as is, but this means changing the precision of the logging mechanism changes the result of the simulation: arguably this will be a very minor change, but the idea is still weird. I would rather remove this mechanism: either always synchronized at 1us as an arbitrary precision threshold (given that having timestamps that vary at a 1e-9 level is not really relevant for robotics), or remove this altogether. I think you did this for unit tests, in which case we can have unit test run with a 1e-8 precision for time without this mechanism and see if that works. 

What do you think?

duburcqa: Good point. I think it is still interesting to sync the timestep with 1us as an arbitrary precision. By doing this, the simulation is not affected by the logging time precision,  time logging at the default precision is more accurate, people using higher accuracy should not really see any difference, and it feels good to know there is a bit of order in the big mess of the timesteps x)

core/include/jiminy/core/telemetry/TelemetryData.h
```diff
@@ -20,12 +20,11 @@ namespace jiminy
     std::string const NUM_INTS("NumIntEntries=");        ///< Number of integers in the data section.
     std::string const NUM_FLOATS("NumFloatEntries=");    ///< Number of floats in the data section.
     std::string const GLOBAL_TIME("Global.Time");        ///< Special column
+    std::string const TIME_UNIT("Global.TIME_UNIT");          ///< Special constant
     std::string const START_CONSTANTS("StartConstants"); ///< Marker of the beginning the constants section.
     std::string const START_COLUMNS("StartColumns");     ///< Marker of the beginning the columns section.
     std::string const START_LINE_TOKEN("StartLine");     ///< Marker of the beginning of a line of data.
     std::string const START_DATA("StartData");           ///< Marker of the beginning of the data section.
-    std::string const START_INPUT_FRAME("INPUT_START");  ///< Marker of the beginning of a input ethercat frame.
```
duburcqa, line 27: Nice catch x)

core/src/telemetry/TelemetryRecorder.cc
```diff
@@ -253,6 +258,21 @@ namespace jiminy
                     isReadingHeaderDone = true;
                 }
 
+                // In header, look for timeUnit constant - if not found, use default time unit.
+                float64_t timeUnit = TELEMETRY_DEFAULT_TIME_UNIT;
+                // Get constants
+                int32_t const lastConstantIdx = std::distance(
```
duburcqa, line 264: you should have a look to `std::find_if`:

```
auto const lastConstantIt = std::find(header.begin(), header.end(), START_COLUMNS);
auto timeUnitIt = std::find_if(header.begin(), lastConstantIt,
                                            [](auto const & constant)
                                            {
                                                 int32_t const delimiter = header[i].find("=");
                                                 return (header[i].substr(0, delimiter) == TIME_UNIT);
                                            });
        if (timeUnitIt != lastConstantIt)
        {
            int32_t const delimiter = timeUnitIt.find("=");
            timeUnit = std::stof(timeUnitIt.substr(delimiter + 1));
        }
```

Or even simpler,

```
auto const lastConstantIt = std::find(header.begin(), header.end(), START_COLUMNS);
for (auto const constantIt = header.begin() ; constantIt != lastConstantIt ; constantIt++)
{
    int32_t const delimiter = constantIt.find("=");
    if (constantIt.substr(0, delimiter) == TIME_UNIT)
    {
        timeUnit = std::stof(constantIt.substr(delimiter + 1));
        break;
    }
}
```

core/src/engine/EngineMultiRobot.cc
matthieuvigne: Totally agree. Even though 1us is arbitrary, from a robotics stand point it really makes sense.

matthieuvigne:
> Ideally, a new terminal condition should be added to the engine before merging:

Done

core/src/telemetry/TelemetryRecorder.cc
```diff
@@ -253,6 +263,20 @@ namespace jiminy
...
+                {
+                    int32_t const delimiter = constantIt->find("=");
+                    if (constantIt->substr(0, delimiter) == TIME_UNIT)
+                    {
+                        timeUnit = std::stof(constantIt->substr(delimiter + 1));
+                        break;
+                    }
+                }
+
+
```
duburcqa, line 279: extra blank line ?

core/include/jiminy/core/telemetry/TelemetryRecorder.h
```diff
@@ -29,11 +29,19 @@ namespace jiminy
...
         ////////////////////////////////////////////////////////////////////////
-        hresult_t initialize(TelemetryData const * telemetryData);
+        hresult_t initialize(TelemetryData * telemetryData,
+                             float64_t const& timeLoggingPrecision);
 
         bool_t const & getIsInitialized(void);
 
+        /// \brief Get the maximum time that can be logged with the current precision.
+        /// \return Max time, in second.
+        float64_t getMaximumLogTime() const;
```
duburcqa, line 43: `float64_t const &` for consistency, even though I agree it does really makes sense for native types.

core/src/engine/EngineMultiRobot.cc
```diff
@@ -808,6 +808,15 @@ namespace jiminy
             returnCode = start(xInit, true, false);
         }
 
+        // Now that telemetry has been initialized, check simulation duration.
+        if (tEnd > telemetryRecorder_->getMaximumLogTime())
```
duburcqa, line 812: You should also check that condition in `step` method.

duburcqa: On the one hand, it would be nice to only have it at one place (in `step`), but on the other hand, it is useful to be warn the user that overflow will happen calling simulate method with too large duration. What do you think ?

duburcqa:
> Done

Nice !

core/src/engine/EngineMultiRobot.cc
```diff
@@ -941,6 +950,15 @@ namespace jiminy
             tEnd = stepperState_.t + stepSize_true;
             stepperState_.tError = (tEnd - stepperState_.t) - stepSize_true;
 
+            // Check that tEnd is not too large for the current logging precision.
+            if (tEnd > telemetryRecorder_->getMaximumLogTime())
```
duburcqa, line 954: line 951 `stepperState_.tError = (tEnd - stepperState_.t) - stepSize_true;`  must be moved after this block.

core/src/engine/EngineMultiRobot.cc
```diff
@@ -1115,9 +1133,8 @@ namespace jiminy
                     while (tNext - t > EPS)
                     {
                         /* Adjust stepsize to end up exactly at the next breakpoint,
-                           prevent steps larger than dtMax, and make sure that dt is
-                           multiple of TELEMETRY_TIME_DISCRETIZATION_FACTOR whenever
-                           it is possible, to reduce rounding errors of logged data. */
+                           prevent steps larger than dtMax, trying to reach multiples of
+                           STEPPER_MIN_TIMESTEP whenever possible. */
```
duburcqa, line 1137: Perhaps we should mention here the pros and cons with discussed ?

core/src/telemetry/TelemetryRecorder.cc
```diff
@@ -87,6 +92,11 @@ namespace jiminy
         return returnCode;
     }
 
+    float64_t TelemetryRecorder::getMaximumLogTime() const
```
duburcqa, line 95: `(void)` for consistency

PR #124, opened 2020-06-03
Fix imu signal

matthieuvigne:
Fix IMU signal computation:
 - spatial acceleration was used instead of linear acceleration.
 - we were missing one forwardKinematics step.

### DIFF
--- a/core/src/robot/Robot.cc
+++ b/core/src/robot/Robot.cc
@@ -1323,6 +1323,10 @@ namespace jiminy
                                Eigen::Ref<vectorN_t const> const & a,
                                vectorN_t                   const & u)
     {
+        // Update kinematic quantities before updating sensors.
+        // There is no need to update frame placement as this has already been done
+        // before.
+        pinocchio::forwardKinematics(pncModel_, pncData_, q, v, a);
         for (auto const & sensorGroup : sensorsGroupHolder_)
         {
             if (!sensorGroup.second.empty())
--- a/core/src/robot/Robot.cc
+++ b/core/src/robot/Robot.cc
@@ -1324,9 +1324,9 @@ namespace jiminy
                                vectorN_t                   const & u)
     {
         // Update kinematic quantities before updating sensors.
-        // There is no need to update frame placement as this has already been done
-        // before.
         pinocchio::forwardKinematics(pncModel_, pncData_, q, v, a);
+        pinocchio::updateFramePlacements(pncModel_, pncData_);
+
         for (auto const & sensorGroup : sensorsGroupHolder_)
         {
             if (!sensorGroup.second.empty())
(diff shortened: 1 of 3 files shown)


core/src/robot/Robot.cc
```diff
@@ -1323,6 +1323,10 @@ namespace jiminy
                                Eigen::Ref<vectorN_t const> const & a,
                                vectorN_t                   const & u)
     {
+        // Update kinematic quantities before updating sensors.
+        // There is no need to update frame placement as this has already been done
+        // before.
+        pinocchio::forwardKinematics(pncModel_, pncData_, q, v, a);
```
matthieuvigne, line 1329: I didn't spend too much time thinking this though and simply added a forwardKinematics step here.

Indeed, even though for an unconstrain simulation we may hope to mutalize with previous computation, when using constraints we call last forwardKinematics(q, v, 0). So we need to set the state back to (q, v, a) at one point: that way, we are sure things are fine when computing the sensor state.

duburcqa: It's fine. Too much optimization kills maintainability and flexibility. But comparing with `computeForwardKinematics`, why are you not calling `updateFramePlacements` ?

unit_py/test_simple_pendulum.py
```diff
@@ -118,6 +118,68 @@ def dynamics(t, x):
         # Compare the numerical and numerical integration of analytical model using scipy
         self.assertTrue(np.allclose(x_jiminy, x_rk_python, atol=TOLERANCE))
 
+    def test_imu_sensor(self):
+        """
+        @brief   Test IMU sensor on pendulum motion.
+
+        @details Since we don't have a simple analytical expression for the solution
```
duburcqa, line 125: Why not to use the solution of the simulation done by jiminy itself ? We are expected the IMU data to be consistent with the solution of Jiminy, not the one computed using an external solver. In my opinion your current implementation is checking at the same time that the solution of the simulation is the right one, and that the sensor IMU data are valid, not just that the IMU data are valid.

core/src/robot/Robot.cc
matthieuvigne: >  why are you not calling `updateFramePlacements`

It's not necessary, but you are right, adding it for clarity. This function doesn't cost much anyway (contrary to framesForwardKinematics which I had in mind, update does not perform any kinematic computation).

unit_py/test_simple_pendulum.py
matthieuvigne: Two reasons:

 A - I like a standalone test better. Though from a software engineer point of view it's redundant (and thus harder to maintain), we can look at the test as a whole and say 'I validated that an IMU mounted on a pendulum gives the signal one would expect from an IMU on a pendulum).

 B - jiminy output log does not feature the acceleration - that way I'm indirectly checking the acceleration computed by jiminy. Though the test could of course be rewritten to compute the expected IMU signal at each time step, using the internal `engine.stepper.dxdt` variable.

duburcqa: No problem! Could you mention that in the test description as a reminder ?

PR #126, opened 2020-06-06
Sensor skew measurement rework

duburcqa:
- Properly sensor initialization in case of delay
- Implement more realistic bias/noise for IMU quaternion since it does not make sense to use simple additive noise/bias with 4 independent coefficients
- Check the size of the user-specified bias and std at setOptions level for the IMU
- Add unit test for sensor skew measurement

### DIFF
--- a/core/src/robot/BasicSensors.cc
+++ b/core/src/robot/BasicSensors.cc
@@ -129,27 +129,30 @@ namespace jiminy
         // Add bias
         if (baseSensorOptions_->bias.size())
         {
-            // Accel + gyroscope: simply add additive bias.
-            get().tail<6>() += baseSensorOptions_->bias.tail<6>();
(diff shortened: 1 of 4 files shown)


core/src/robot/BasicSensors.cc
```diff
@@ -102,25 +102,40 @@ namespace jiminy
...
         data().tail<3>() = acceleration.linear() +
                            velocity.angular().cross(velocity.linear()) -
                            quat.conjugate() * robot_->pncModel_.gravity.linear();
 
+        return hresult_t::SUCCESS;
+    }
 
+    void ImuSensor::skewMeasurement(void)
+    {
+        // Add white noise
```
duburcqa, line 128: @matthieuvigne I need you for that stuff !

PR #162, opened 2020-08-09
[Win10] Update Jiminy dependencies versions

duburcqa:
- Applying patch files (reduced to bare minimum) on dependencies instead of manual source modifications
- Dependency version parity between Windows and Linux
- Get rid of pkgconfig completely for both Windows and Linux
- Move to C++14 for every dependencies

Note that some methods of `bindings\python\algorithm\expose-joints.cpp` have been disabled to reduce memory heap consumption during compilation (from 34Go down to about 2Go), namely `integrate`, `dIntegrate`, `difference`, and `dDifference` and `interpolate`. Whose methods are usually not critical for Python bindings.

It also fixes some minor bugs introduced by the previous PR.


python/jiminy_py/src/jiminy_py/viewer.py
```diff
@@ -551,6 +551,7 @@ def close(self=None):
             if self is None:
                 self = Viewer
             else:
+                Viewer._backend_robot_names.remove(self.robot_name)
```
duburcqa, line 554: add comment "Consider that the robot is now available, no matter whether the robot has actually been deleted or not"

duburcqa:
The option  `close_backend_at_exit=True,` is unsafe and should be removed.

build_tools/build_install_deps_windows.ps1
```diff
@@ -74,41 +82,7 @@ git submodule --quiet update --init --recursive --jobs 8
...
-#   pragma pop_macro("xor")
-#endif
-#endif
-#endif'
-} ; $_ ; $n++} | `
-Out-File -Encoding ASCII "$RootDir/boost/libs/python/include/boost/python/operators.hpp"
-Set-PSDebug -Trace 1
-
 ### Build and install the build tool b2 (build-ception !)
-Set-Location -Path "$RootDir/boost"
```
duburcqa, line 111: This Line shouldn't be removed.

duburcqa:
> The option close_backend_at_exit is unsafe and should be removed completely.

Done

PR #163, opened 2020-08-10
Add record trajectories option

RoucherA:
-  Add option to record trajectories in viewer
- Fix robot deletion issue in viewer.
- Increase velocity bounds for gym obs space to make sure it never goes out-of-bounds


python/jiminy_py/src/jiminy_py/viewer.py
```diff
@@ -11,6 +11,8 @@
 import signal
 import base64
 import atexit
+import cv2
```
duburcqa, line 14: Missing dependency in setup.py

PR #196, opened 2020-09-09
[Gym] Generic locomotion envs - PART II

duburcqa:
- Enable to compute collision on convex hull of collision meshes.
- Bind collision model/data in Python to allow fetching contact information
- Clear distinction between usual contact points and collision bodies
- Use same, latest, Jiminy's dependencies on Ubuntu 18 and Ubuntu 20.


duburcqa:
The order of the geometry object in the collision pairs définition is important because it determines the direction of the normal. It seems that the ground should be the first object to get the normal of the body wrt the world, since the normal of the ground wrt the world is constant and not interesting. Here are some related issues [1](https://github.com/flexible-collision-library/fcl/issues/8#issuecomment-93856329), [2](https://github.com/flexible-collision-library/fcl/pull/61#issuecomment-203955114) on fcl repo.

duburcqa:
Also, it would be better to compute at least 3 contact points per body to improve the stability of the contact modeling. It can be done pretty easily and selectively for each body by setting the attribute std::vector `CollisionRequests` of GeometryData. 
Maybe this number of contact points should be in input argument of addCollisionBodies, but it is not clear how to do it since it may ne different for each bodies. Perhaps it could also be a config option of the model.

PR #205, opened 2020-10-03
[core] Implement Lie-group based steppers.

matthieuvigne:
Implement custom steppers working in Lie group directly, instead of relying on boost for vector-space steppers.
This gives a significant improvement to accuracy and performance for some systems with quaternions.


core/include/jiminy/core/stepper/AbstractStepper.h
```diff
@@ -0,0 +1,82 @@
...
+#define JIMINY_ABSTRACT_STEPPER_H
+
+#include <functional>
+
+#include "jiminy/core/Types.h"
+
+namespace jiminy
+{
+    class Robot;
+    struct systemState_t;
```
duburcqa: Useless.

core/include/jiminy/core/stepper/AbstractStepper.h
```diff
@@ -0,0 +1,82 @@
...
+
+    template<typename F1>
+    class AbstractStepper
+    {
+        public:
+            /// \brief Constructor
+            /// \param[in] f      Dynamics function, with signature a = f(t, q, v)
+            /// \param[in] robots Robots whose dynamics the stepper will work on.
+            AbstractStepper(F1 & f,
```
duburcqa: Personally I like `(robots, f)` more, but as you want.

core/include/jiminy/core/stepper/AbstractStepper.h
```diff
@@ -0,0 +1,82 @@
...
+            ///          by variable step schemes to indicate the next advised dt.
+            ///
+            /// \param[in]      t     Integration start time.
+            /// \param[in, out] q     System starting position.
+            /// \param[in, out] v     System starting velocity.
+            /// \param[in, out] a     System starting acceleration.
+            /// \param[in, out] dt    Input: desired integration duration. Output: recommanded step size for
+            ///                       variable-step schemes. Constant-step schemes leave this value unmodified.
+            /// \return True if integration was successful, false otherwise. In that case, (q, v, a) are not updated.
+            virtual bool try_step(float64_t const& t,
```
duburcqa: rather `(q, v, a, t, dt)` for "consistency with boost ?

core/include/jiminy/core/stepper/AbstractStepper.h
```diff
@@ -0,0 +1,82 @@
...
+                                  std::vector<vectorN_t> & q,
+                                  std::vector<vectorN_t> & v,
+                                  std::vector<vectorN_t> & a,
+                                  float64_t & dt) = 0;
+
+        private:
```
duburcqa: `protected`, to be accessible by subclasses

core/include/jiminy/core/stepper/AbstractStepper.h
```diff
@@ -0,0 +1,82 @@
...
+
+            /// \brief Increment the given velocity by the given acceleration for all system.
+            /// \details addToVelocity(v, a * dt) simply computes v + a * dt for all systems.
+            /// \param[in] velocityIn Starting velocity
+            /// \param[in] accelerationIn Desired acceleration increment
+            /// \return Incremented velocity
+            void addToVelocity(std::vector<vectorN_t> const & velocityIn, std::vector<vectorN_t> const& accelerationIn);
+
+
+            std::conditional_t<std::is_function<F1>::value,
```
duburcqa: `private`

core/include/jiminy/core/stepper/AbstractStepper.tpp
```diff
@@ -0,0 +1,50 @@
...
+
+    template<typename F1>
+    std::vector<vectorN_t> AbstractStepper<F1>::addToPosition(
+        std::vector<vectorN_t> const & positionIn,
+        std::vector<vectorN_t> const& velocityIn)
+    {
+        assert(positionIn.size() == robots_.size())
+        assert(velocityIn.size() == robots_.size())
+
+        std::vector<vectorN_t> positionOut;
```
duburcqa: `positionOut` in input argument, and internal buffers for each knots to avoid memory allocation.

core/include/jiminy/core/stepper/AbstractStepper.tpp
```diff
@@ -0,0 +1,50 @@
...
+        for (int i = 0; i < robots_.size(); i++)
+        {
+            vectorN_t position = positionIn.at(i);
```
duburcqa: `[i]` instead of `at` to avoid checking size systematically (they are not equivalent, `at` is slower)

core/include/jiminy/core/stepper/AbstractStepper.tpp
```diff
@@ -0,0 +1,50 @@
...
+
+    template<typename F1>
+    std::vector<vectorN_t> AbstractStepper<F1>::addToVelocity(
+        std::vector<vectorN_t> const & velocityIn,
+        std::vector<vectorN_t> const& accelerationIn)
+    {
+        assert(velocityIn.size() == robots_.size())
+        assert(accelerationIn.size() == robots_.size())
+
+        std::vector<vectorN_t> velocityOut;
```
duburcqa: Same

core/include/jiminy/core/stepper/AbstractStepper.h
matthieuvigne: Yes and I don't know why. Functions are always written as `f(t, x)`, yet for some reasons numerical schemes are `scheme(f, x, t)``` - this is the case for boost and scipy for instance. I think it's because `t` is optional for autonomous dynamics (but then why isn't the same convention used for `f`.

Here I liked `f(t, x, dt)` better also because t is constant, while all following arguments are modified as reference: it fells weird to have a const in the middle of return values...

core/include/jiminy/core/stepper/AbstractStepper.h
matthieuvigne: f_ needs to be accessed by sub-classes as well - not robots_.

duburcqa: indeed

core/include/jiminy/core/stepper/AbstractStepper.h
duburcqa: No problem !

core/include/jiminy/core/stepper/AbstractStepper.tpp
matthieuvigne: Ah I see what you want to do.  In that case I think the best API is `add(q0, v, q1)` because  for Runge-Kutta because I have to do several times `qi = q0 + vi` with the same q0.
Won't there still be memory allocation though due to the fact that vectorN_t are not fixed sized ? How is `stdVect[i] = otherStdVect[i]``` handled, given that both Eigen::Vectors may have different sizes ? Should I enforce fixed size from the start.

matthieuvigne: Also there is the question of doing a sum of more than one element: I could use successive addPosition() but this might not be very efficient... 
I think it would make sense to write it rather as a function with a variable number of arguments, which should be possible with fancy C++:

```
addPosition(qOut, q0, lambdai, q1, ... lambdaN qN): 
qOut = q0 + sum_i lambdai qi

```

matthieuvigne: The problem is that I'm mixing types between lambda (scalar) and q (std::vector).
I could make use of std::pair, but this is likely to induce an unwanted overhead.

duburcqa: > In that case I think the best API is add(q0, v, q1)

I agree.

> Won't there still be memory allocation though due to the fact that vectorN_t are not fixed sized ? How is `stdVect[i] = otherStdVect[i]``` handled, given that both Eigen::Vectors may have different sizes ? Should I enforce fixed size from the start.

You need one set of buffers for each system. So you need should define a `systemBuffers_t` struct to store them for each of them.

> I think it would make sense to write it rather as a function with a variable number of arguments, which should be possible with fancy C++:

Yes nice idea ! Just add a TODO and the method declaration and I could implement it using variadic template.

matthieuvigne: Can variadic templates handle the fact that the signature is (type1, type2, type1, type2...) without creating a pair<type1, type2> ?

duburcqa: Yes of course it is possible ! The trick is to template the method with any number of arguments of any types, then to use SFINAE to check at compile time that the arguments are appropriate using "traits" and std::enable_if. Easy !

```
#include <type_traits>

struct A {};
struct B {};

template <typename X, typename Y, typename... Ts>
struct alternates : std::false_type {};

template <typename X, typename Y>
struct alternates<X, Y> : std::true_type {};

template <typename X, typename Y, typename... Ts>
struct alternates<X, Y, X, Ts...>
: std::conditional<alternates<Y, X, Ts...>::value, std::true_type, std::false_type>::type {};

struct C
{
    template<typename... Ts,
        typename = typename std::enable_if<
            sizeof...(Ts) % 2 == 0 &&
            sizeof...(Ts) >= 2 &&
            alternates<A, B, Ts...>::value
        >::type>
    C(Ts&&...);
};

int main() {
    C c1 { A(), B()};
    C c2 { A(), B(), A(), B(), A(), B() };
    // C c3 {};                     // I assumed you need at least 2
    // C c4 { A(), B(), A(), A() }; // A, A doesn't alternate
    // C c5 { B(), A(), B() };      // B, A, B not allowed
    // C c6 { A(), B(), A(), B() }; // A, B, A, B doesn't pair
}
```

core/include/jiminy/core/stepper/AbstractStepper.h
```diff
@@ -0,0 +1,82 @@
...
+
+#include "jiminy/core/Types.h"
+
+namespace jiminy
+{
+    class Robot;
+    struct systemState_t;
+
+    template<typename F1>
+    class AbstractStepper
```
duburcqa, line 32: Sorry but I forgot that an abstract must NOT be templated (because the type is always unknown in advance in practice), so here we must hind the actual functor type using a std::function:


```
    using systemDynamics = std::function<void(float64_t const & /*t*/,
                                              std::vector<vectorN_t> const & /*qSplit*/,
                                              std::vector<vectorN_t> const & /*vSplit*/,
                                              std::vector<vectorN_t> & /*aSplit*/)>;
```

core/include/jiminy/core/stepper/LieGroup.h
```diff
@@ -0,0 +1,82 @@
...
+#ifndef JIMINY_LIE_GROUP_H
+#define JIMINY_LIE_GROUP_H
+
+#include "jiminy/core/Types.h"
+
+namespace jiminy
+{
+    class Robot;
+
+    class State
```
duburcqa: You need to forward-declare `StateDerivative` before defining `State` class since you are declaring `+` operator, which requires knowing this type exists.

duburcqa: I would rather use a struct, to emphasize that it is not a "real" class but rather a way to identify a data structure with specific operators.

PR #226, opened 2020-11-10
[misc] Use static linkage for every Boost components except Python

duburcqa:


duburcqa:
Related to https://discuss.python.org/t/delocate-auditwheel-but-for-windows/2589/9

PR #233, opened 2020-11-27
[core] Improve joints position/velocity bounds enforcement.

duburcqa:


core/src/engine/EngineMultiRobot.cc
```diff
@@ -2118,6 +2126,188 @@ namespace jiminy
...
+            else
+            {
+                return;
+            }
+
+            // Generate acceleration in the opposite direction if out-of-bounds
+            float64_t const accelJoint = - 2.0 * jointOptions.boundDamping * vJointError;
+
+            // Apply the resulting force
+            u[velocityIdx] += Ia * accelJoint;
```
jcarpent, line 2290: Small comment: this is for sure something you don't want to exploit on real robots. RL will do, for sure.

duburcqa: Are you talking about trying to enforce position/velocity bounds, or the approach to implement them ?
And by "exploit on real robots", you mean for online computation on real hardware ? Because for sure it is not the intended use-case. Maybe it should be disable by default.

jcarpent: The thing is that both optimization and RL  approaches will try to exploit this artefact in order to provide impulses on the system (because they will act as magic forces to control your dynamics). But in general, you don't want to hit the joint limits (otherwise, you will break your robots). If you model it, then there is a chance to get the counter effects, meaning enforcing the constraints to exploit it.

duburcqa: Ah yes I get your point, and I agree with you. Depending on the application, it may actually be helpful to speed up learning and avoid gathering observation that are impossible in practice. But clearly it is likely to do more harm than good in many cases.

PR #263, opened 2021-01-09
More generic flexibilities and fix external forces computation

duburcqa:


core/include/jiminy/core/engine/PinocchioOverloadAlgorithms.h
```diff
@@ -57,6 +60,32 @@ namespace pinocchio_overload
         return data.tau;
     }
 
+    template<typename Scalar, int Options, template<typename, int> class JointCollectionTpl,
+             typename ConfigVectorType, typename TangentVectorType1, typename TangentVectorType2>
+    inline const typename pinocchio::DataTpl<Scalar,Options,JointCollectionTpl>::TangentVectorType &
+    rnea(pinocchio::ModelTpl<Scalar, Options, JointCollectionTpl> const & model,
```
matthieuvigne, line 66: Why is this part of this MR ?

core/src/engine/EngineMultiRobot.cc
```diff
@@ -2705,37 +2712,42 @@ namespace jiminy
...
+            a = pinocchio_overload::aba(
                 system.robot->pncModel_, system.robot->pncData_, q, v, u, fext);
         }
+
+        // Compute the internal forces
+        pinocchio_overload::rnea(
+            system.robot->pncModel_, system.robot->pncData_, q, v, a);
+
+        // Using action-reaction law to compute the ground reaction force
+        system.robot->pncData_.f[0].setZero();
```
matthieuvigne, line 2739: I'm not sure to understand why you need this still. This would only have an impact if you run rnea an extra time, but isn't the first run enough ?

duburcqa: This is not updated nor set to zero by rnea. Si it must be done manually.

core/include/jiminy/core/engine/PinocchioOverloadAlgorithms.h
duburcqa: What do you mean ? We need to be able to call rnea without external forces to compute the sum of the forces applied on each body, which was not the case up to this day.

PR #279, opened 2021-02-02
[misc] Add wheels for Python3.9.

duburcqa:


duburcqa:
For now, it is impossible to support Python 3.9, since it relies on `Pillow` <= 7.2.0 (because `gym` does...), which is not compatible ( >=  8.0 is the minimum version compatible) and numba, for which no version compatible exists at all.

duburcqa:
numba 0.53 will support Python3.9 and should be released anytime soon (rc3 is already available).

PR #284, opened 2021-02-28
[core] Add force impulse contact model.

### LINKED ISSUE #47: [docs|core] Standard contact/friction model
It would be nice to use a standard model for contact and friction if it is not the use, but more importantly to document it.

duburcqa:


core/src/robot/LCPSolvers.cc
```diff
@@ -0,0 +1,221 @@
...
+        matrixN_t & A = data.JMinvJt;
+        vectorN_t & f = data.lambda_c;
+
+        // Keep previous value if size did not change, reset it otherwise
+        if (f.size() != gamma.size())
+        {
+            f = vectorN_t::Zero(gamma.size());
+        }
+
+        // Compute the UDUt decomposition of data.M
```
jcarpent, line 161: Please keep the copyright holders in all the files where you have done copy/pasting. This is a legal issue!

jcarpent (changes requested):

core/src/robot/LCPSolvers.cc
duburcqa: Is it ok now ? https://github.com/Wandercraft/jiminy/blob/afd2a84af5a038e641ffd713af9a5e05bd7826ed/core/src/robot/LCPSolvers.cc#L140

jcarpent: Thanks for accounting for this request.
I think it also does apply for DART if the code you mention is originally coming from this framework ;)

duburcqa: In the particular case of DART, the rework is more significant. I use modern C++17 stl algorithms, Eigen-based linear algebra instead of custom operations on raw C flattened array pointers, and a small mathematical trick to speed-up computations by leveraging CPU vectorization. Both implementation are mathematically equivalent yet the implement is different (no copy-paste at all).

jcarpent: If you say so, I won't check it. 
But as you're working for a company, it corresponds to some classic and standard legal issues I would say.

duburcqa: What do you mean ? I should double check that I'm done violating any legal stuff ? Jiminy is not related to any company in practice, and I'm thinking to completely get rid of Wandercraft repository, which is basically a mirror of my own repository. Do you think it could help ?

jcarpent: Let's discuss it on private channels for legal issues if you want. I let you contact me.

PR #290, opened 2021-03-08
[jiminy/viewer] Panda3d has default viewer backend.

duburcqa:


duburcqa:
```
            # Replace non-standard hard drive prefix on Windows
            import os
            import re
            import sys
            if sys.platform.startswith('win'):
                mesh_path = re.sub(r'^/([A-Za-z])', lambda m: m.group(1).upper() + ":", mesh_path)  # Fix hard drive prefix
```

```
            # Assimp backend used to load meshes does not support many things related to
            # paths on Windows. First, it does not support symbolinks, then the hard drive
            # prefix must be `/x/` instead of `X:\`, and finally backslashes must be used 
            # as delimiter instead of forwardslashes.
            import os
            import re
            import sys
            from pathlib import PureWindowsPath
            mesh_path = geometry_object.meshPath
            if sys.platform.startswith('win'):
                mesh_path = os.path.realpath(mesh_path) # Resolve symlinks
                mesh_path = PureWindowsPath(mesh_path).as_posix()  # Replace delimiters
                mesh_path = re.sub(r'^([A-Za-z]):', lambda m: "/" + m.group(1).lower(), mesh_path)  # Fix hard drive prefix
```

```
        # Check that the final posture matches the expected one.
        robot_name = self.env.robot.pinocchio_model.name
        i = 0
        img_diff = np.inf
        while img_diff > 0.1:
            img_name = '_'.join((
                robot_name, f"standing_{self.env.viewer.backend}_{i}.png"))
            data_dir = os.path.join(os.path.dirname(__file__), "data")
            img_fullpath = os.path.join(data_dir, img_name)
            # plt.imsave(img_fullpath, rgb_array)
            try:
                rgba_array_rel_orig = plt.imread(img_fullpath)
            except FileNotFoundError:
                break
            rgb_array_abs_orig = (
                rgba_array_rel_orig[..., :3] * 255).astype(np.uint8)
            img_diff = np.mean(np.abs(rgb_array - rgb_array_abs_orig))
            i += 1
        self.assertTrue(img_diff < 0.1)
```

atlas_standing_panda3d_0
atlas_standing_panda3d_1

PR #432, opened 2021-10-29
WIP: [core] Add transmission abstraction.

### LINKED ISSUE #428: [core] Add support of motors to joints transmissions

fabinsch:
start modifying `AbstractMotor` and `BasicMotors`, add initial files for transmission


fabinsch:
Todo:
- [ ]  energy dissipation in transmission
- [ ]  transmission constraints

core/include/jiminy/core/robot/AbstractMotor.h
```diff
@@ -132,84 +136,65 @@ namespace jiminy
...
 
         ///////////////////////////////////////////////////////////////////////////////////////////////
-        /// \brief      Get jointPositionIdx_.
+        /// \brief      Set the same configuration options for every motors.
         ///
-        /// \details    It is the index of the joint associated with the motor in the configuration vector.
+        /// \param[in]  motorOptions   Dictionary with the parameters used for any motor
         ///////////////////////////////////////////////////////////////////////////////////////////////
-        int32_t const & getJointPositionIdx(void) const;
+        hresult_t setOptionsAll(configHolder_t const & motorOptions);
```
duburcqa: I suggest keeping the function that you did not modify at the same place in the file, so that it is not part of the diff. It would make the review easier. For instance, this method was still there before and yet is part of both deletion and addition.

core/include/jiminy/core/robot/AbstractMotor.h
```diff
@@ -253,39 +230,49 @@ namespace jiminy
...
+        float64_t & q(void);
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Get a reference to the last data buffer corresponding to the actual velocity
+        ///             of the motor.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        float64_t & v(void);
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Get a reference to the last data buffer corresponding to the actual acc
```
duburcqa: acceleration.

core/include/jiminy/core/robot/AbstractMotor.h
```diff
@@ -253,39 +230,49 @@ namespace jiminy
...
+        ///             of the motor.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        float64_t & a(void);
+
         ///////////////////////////////////////////////////////////////////////////////////////////////
         /// \brief      Get a reference to the last data buffer corresponding to the actual effort
         ///             of the motor.
         ///////////////////////////////////////////////////////////////////////////////////////////////
-        float64_t & data(void);
+        float64_t & u(void);
```
duburcqa: configure your editor to automatically remove extra white spaces. If it is VSCode:
```
"files.trimTrailingWhitespace": true,
```

core/include/jiminy/core/robot/AbstractMotor.h
```diff
@@ -253,39 +230,49 @@ namespace jiminy
...
         ///////////////////////////////////////////////////////////////////////////////////////////////
-        /// \brief    Attach the sensor to a robot
+        /// \brief    Attach the motor to a robot
         ///
-        /// \details  This method must be called before initializing the sensor.
+        /// \details  This method must be called before initializing the motor.
         ///////////////////////////////////////////////////////////////////////////////////////////////
         hresult_t attach(std::weak_ptr<Robot const> robot,
-                         std::function<hresult_t(AbstractMotorBase & /*motor*/)> notifyRobot,
+                         std::function<hresult_t(AbstractMotorBase & /* motor */)> notifyRobot,
```
duburcqa: you still need `notifyRobot` ?

core/include/jiminy/core/robot/AbstractMotor.h
```diff
@@ -294,19 +281,13 @@ namespace jiminy
 
     protected:
         configHolder_t motorOptionsHolder_;                          ///< Dictionary with the parameters of the motor
-        bool_t isInitialized_;                                       ///< Flag to determine whether the controller has been initialized or not
         bool_t isAttached_;                                          ///< Flag to determine whether the motor is attached to a robot
         std::weak_ptr<Robot const> robot_;                           ///< Robot for which the command and internal dynamics
-        std::function<hresult_t(AbstractMotorBase &)> notifyRobot_;  ///< Notify the robot that the configuration of the sensors have changed
+        std::function<hresult_t(AbstractMotorBase &)> notifyRobot_;  ///< Notify the robot that the configuration of the motors have changed
```
duburcqa, line 295: same.

core/include/jiminy/core/robot/AbstractMotor.h
```diff
@@ -294,19 +281,13 @@ namespace jiminy
...
         std::string name_;                                           ///< Name of the motor
         int32_t motorIdx_;                                           ///< Index of the motor in the measurement buffer
-        std::string jointName_;
-        jointIndex_t jointModelIdx_;
-        joint_t jointType_;
-        int32_t jointPositionIdx_;
-        int32_t jointVelocityIdx_;
         float64_t commandLimit_;
-        float64_t armature_;
+        float64_t armature_;
```
duburcqa: white space. Also, you should review yourself your diff to make sure to are not adding meaningless diff that could be avoided using `git add -p .` systematically.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,260 @@
...
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Get the actual state of the transmission at the current time.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        float64_t const & get(void) const;
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Get the actual state of all the transmissions at the current time.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        vectorN_t const & getAll(void) const;
```
duburcqa: This method is irrelevant for transmissions.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,260 @@
...
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & q,
+                                             Eigen::VectorBlock<vectorN_t const> const & v) = 0;  /* copy on purpose */
+
+        
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Compute energy dissipation in the transmission.
+        ///
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual void computeEffortTransmission() = 0;
```
duburcqa: `computeEffortTransmission(void)` by convention in jiminy.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,260 @@
...
+
+    private:
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief    Attach the transmission to a robot
+        ///
+        /// \details  This method must be called before initializing the transmission.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        hresult_t attach(std::weak_ptr<Robot const> robot,
+                         std::function<hresult_t(AbstractTransmissionBase & /*transmission*/)> notifyRobot);
```
duburcqa: you need `notifyRobot` ?

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,260 @@
...
+
+    public:
+        std::unique_ptr<abstractTransmissionOptions_t const> baseTransmissionOptions_;  ///< Structure with the parameters of the transmission
+
+    protected:
+        configHolder_t transmissionOptionsHolder_;                   ///< Dictionary with the parameters of the transmission
+        bool_t isInitialized_;                                       ///< Flag to determine whether the transmission has been initialized or not
+        bool_t isAttached_;                                          ///< Flag to determine whether the transmission is attached to a robot
+        std::weak_ptr<Robot const> robot_;                           ///< Robot for which the command and internal dynamics
+        std::function<hresult_t(AbstractTransmissionBase &)> notifyRobot_;  ///< Notify the robot that the configuration of the transmissions have changed
```
duburcqa: same.

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -0,0 +1,61 @@
...
+        auto shared_from_this() { return shared_from(this); }
+        auto shared_from_this() const { return shared_from(this); }
+
+        hresult_t initialize(std::string const & jointName,
+                             std::string const & motorName);
+
+        virtual hresult_t setOptions(configHolder_t const & transmissionOptions) final override;
+
+    private:
+        virtual float64_t computeTransform(Eigen::VectorBlock<vectorN_t> /* q */,
```
duburcqa: uncomment variable names in declarations.

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -0,0 +1,61 @@
...
+                                           Eigen::VectorBlock<vectorN_t> /* v */) final override;
+
+        virtual float64_t computeInverseTransform(Eigen::VectorBlock<vectorN_t> /* q */,
+                                                  Eigen::VectorBlock<vectorN_t> /* v */) final override;
```
duburcqa: missing final `;` apparently.

core/include/jiminy/core/robot/Model.h
```diff
@@ -26,14 +27,16 @@ namespace jiminy
         BOUNDS_JOINTS = 0,
         CONTACT_FRAMES = 1,
         COLLISION_BODIES = 2,
-        USER = 3
+        USER = 3,
+        TRANSMISSIONS = 4
```
duburcqa, line 31: Move back `USER` at the end.

core/include/jiminy/core/robot/Robot.h
```diff
@@ -53,6 +55,7 @@ namespace jiminy
         motorsHolder_t const & getMotors(void) const;
         hresult_t detachMotor(std::string const & motorName);
         hresult_t detachMotors(std::vector<std::string> const & motorsNames = {});
+        hresult_t detachTransmissions(std::vector<std::string> const & transmissionsNames = {});
```
duburcqa, line 59: Missing `hresult_t detachTransmission(std::string const & transmissionName);`

core/include/jiminy/core/robot/Robot.h
```diff
@@ -64,6 +67,7 @@ namespace jiminy
         hresult_t detachSensor(std::string const & sensorType,
                               std::string const & sensorName);
         hresult_t detachSensors(std::string const & sensorType = {});
+        transmissionsHolder_t const & getTransmissions(void) const;
```
duburcqa, line 73: missing `hresult_t getTransmission(std::string const & transmissionName, std::shared_ptr<AbstractTransmissionBase> & transmission)`

core/include/jiminy/core/robot/Robot.h
```diff
@@ -123,6 +127,8 @@ namespace jiminy
         std::vector<int32_t> getMotorsVelocityIdx(void) const;
         std::unordered_map<std::string, std::vector<std::string> > const & getSensorsNames(void) const;
         std::vector<std::string> const & getSensorsNames(std::string const & sensorType) const;
+        std::vector<std::string> const & getActuatedJoints(void) const;
+        hresult_t updateActuatedJoints(std::vector<std::string> const & jointNames);
```
duburcqa: WTF this method is supposed to be x) It is not possible to manually update the list of actuated joints! It should be determined based on attached transmissions.

core/include/jiminy/core/robot/Robot.h
```diff
@@ -123,6 +127,8 @@ namespace jiminy
         std::vector<int32_t> getMotorsVelocityIdx(void) const;
         std::unordered_map<std::string, std::vector<std::string> > const & getSensorsNames(void) const;
         std::vector<std::string> const & getSensorsNames(std::string const & sensorType) const;
+        std::vector<std::string> const & getActuatedJoints(void) const;
```
duburcqa: `std::vector<std::string> const & getActuatedJointsNmaes(void) const;`

core/include/jiminy/core/robot/Robot.h
```diff
@@ -145,13 +151,15 @@ namespace jiminy
...
         motorsHolder_t motorsHolder_;
+        transmissionsHolder_t tranmissionsHolder_;
         sensorsGroupHolder_t sensorsGroupHolder_;
         std::unordered_map<std::string, bool_t> sensorTelemetryOptions_;
         std::vector<std::string> motorsNames_;                                      ///< Name of the motors
         std::unordered_map<std::string, std::vector<std::string> > sensorsNames_;   ///< Name of the sensors
         std::vector<std::string> commandFieldnames_;                                ///< Fieldnames of the command
         std::vector<std::string> motorEffortFieldnames_;                            ///< Fieldnames of the motors effort
         uint64_t nmotors_;                                                          ///< The number of motors
+        std::vector<std::string> actuatedJoints_;                                   ///< List of joints attached to a transmission
```
duburcqa: `actuatedJointsNames_`

core/src/robot/AbstractMotor.cc
```diff
@@ -65,9 +60,14 @@ namespace jiminy
         motorIdx_ = sharedHolder_->num_;
 
         // Add a value for the motor to the shared data buffer
-        sharedHolder_->data_.conservativeResize(sharedHolder_->num_ + 1);
-        sharedHolder_->data_.tail<1>().setZero();
-
+        sharedHolder_->position_.conservativeResize(sharedHolder_->num_ + 1);
```
duburcqa: Just to some loops:
```
for (vectorN_t & data : std::array<vectorN_t &, 4>{{
        sharedHolder_->position_,
        sharedHolder_->velocity_,
        sharedHolder_->acceleration_,
        sharedHolder_->effort_}})
{
    [TODO]
}
```

core/src/robot/AbstractMotor.cc
```diff
@@ -92,10 +92,20 @@ namespace jiminy
         if (motorIdx_ < sharedHolder_->num_ - 1)
         {
             int32_t motorShift = sharedHolder_->num_ - motorIdx_ - 1;
-            sharedHolder_->data_.segment(motorIdx_, motorShift) =
-                sharedHolder_->data_.segment(motorIdx_ + 1, motorShift).eval();  // eval to avoid aliasing
+            sharedHolder_->position_.segment(motorIdx_, motorShift) =
```
duburcqa: same.

core/src/robot/AbstractMotor.cc
```diff
@@ -92,10 +92,20 @@ namespace jiminy
...
+            sharedHolder_->velocity_.segment(motorIdx_, motorShift) =
+                sharedHolder_->velocity_.segment(motorIdx_ + 1, motorShift).eval();  // eval to avoid aliasing
+            sharedHolder_->acceleration_.segment(motorIdx_, motorShift) =
+                sharedHolder_->acceleration_.segment(motorIdx_ + 1, motorShift).eval();  // eval to avoid aliasing
+            sharedHolder_->effort_.segment(motorIdx_, motorShift) =
+                sharedHolder_->effort_.segment(motorIdx_ + 1, motorShift).eval();  // eval to avoid aliasing
+                                                
         }
-        sharedHolder_->data_.conservativeResize(sharedHolder_->num_ - 1);
+        sharedHolder_->position_.conservativeResize(sharedHolder_->num_ - 1);
```
duburcqa: same.

core/src/robot/AbstractMotor.cc
```diff
@@ -138,7 +148,10 @@ namespace jiminy
         }
 
         // Clear the shared data buffer
-        sharedHolder_->data_.setZero();
+        sharedHolder_->position_.setZero();
```
duburcqa: same.

core/src/robot/AbstractMotor.cc
duburcqa: I would even suggest to gather all the small loops in one big or two if possible.

core/src/robot/AbstractMotor.cc
```diff
@@ -236,33 +240,12 @@ namespace jiminy
...
-        if (returnCode == hresult_t::SUCCESS)
-        {
-            ::jiminy::getJointPositionIdx(robot->pncModel_, jointName_, jointPositionIdx_);
             ::jiminy::getJointVelocityIdx(robot->pncModel_, jointName_, jointVelocityIdx_);
 
             // Get the motor effort limits from the URDF or the user options.
             if (baseMotorOptions_->commandLimitFromUrdf)
             {
-                commandLimit_ = robot->pncModel_.effortLimit[jointVelocityIdx_] / baseMotorOptions_->mechanicalReduction;
+                commandLimit_ = robot->pncModel_.effortLimit[jointVelocityIdx_];
```
duburcqa: This is wrong. You need to add a function `getMotorEffortLimit` or something if not already available at robot level, because it `pncModel_.effortLimit` is at joint level and it is how it should be, so the information you need is missing. You need to combine robot information and transmission information to get it. So better add a abstract method at robot level to do so !

core/src/robot/BasicMotors.cc
```diff
@@ -79,25 +62,16 @@ namespace jiminy
         return returnCode;
     }
 
-    hresult_t SimpleMotor::computeEffort(float64_t const & /* t */,
-                                         Eigen::VectorBlock<vectorN_t const> const & /* q */,
-                                         float64_t const & v,
-                                         float64_t const & /* a */,
-                                         float64_t command)
+    hresult_t SimpleMotor::computeEffort(float64_t const & v,
+                                        float64_t command)
```
duburcqa: indentation.

core/src/robot/AbstractTransmission.cc
```diff
@@ -0,0 +1,442 @@
...
+    AbstractTransmissionBase::~AbstractTransmissionBase(void)
+    {
+        // Detach the transmission before deleting it if necessary
+        if (isAttached_)
+        {
+            detach();
+        }
+    }
+
+    hresult_t AbstractTransmissionBase::initialize(void)
```
duburcqa: This is wrong, the signature should be `std::string const & jointNames, std::string const & motorNames` already, not void, since you clearly rely on the fact that `jointNames_` and `motorNames_` are going to be set.

core/src/robot/BasicTransmissions.cc
```diff
@@ -0,0 +1,77 @@
...
+    AbstractTransmissionBase(name),
+    transmissionOptions_(nullptr)
+    {
+        /* AbstractTransmissionBase constructor calls the base implementations of
+           the virtual methods since the derived class is not available at
+           this point. Thus it must be called explicitly in the constructor. */
+        setOptions(getDefaultTransmissionOptions());
+    }
+
+    hresult_t SimpleTransmission::initialize(std::string const & jointNames, std::string const & motorNames)
```
duburcqa: The whole implementation should be in abstract implementation, not the specialized one.

core/src/robot/BasicTransmissions.cc
```diff
@@ -0,0 +1,77 @@
...
+        {
+            jointNames_.clear();
+            motorNames_.clear();
+            isInitialized_ = false;
+        }
+
+        return returnCode;
+    }
+
+    hresult_t SimpleTransmission::setOptions(configHolder_t const & transmissionOptions)
```
duburcqa: This is doing basically nothing and there is no custom option for the basic transmission, so just remove this method overload completely.

core/src/robot/Robot.cc
```diff
@@ -705,6 +851,11 @@ namespace jiminy
         return sensorsGroupHolder_;
     }
 
+    Robot::tranmissionssGroupHolder_t const & Robot::getTransmissions(void) const
```
duburcqa: `tranmissionss` spelling mistake.

duburcqa: Just remove this methode, since `tranmissionssGroupHolder_t` does not exist anyway, and the method is already implemented right above.

core/src/robot/Robot.cc
```diff
@@ -1216,6 +1367,12 @@ namespace jiminy
         return motorEffortEmpty;
     }
 
+    hresult_t Robot::updateActuatedJoints(std::vector<std::string> const & jointNames)
```
duburcqa: Remove this method, it does not make sense.

core/src/robot/Robot.cc
```diff
@@ -1407,7 +1564,7 @@ namespace jiminy
 
     vectorN_t const & Robot::getArmatures(void) const
     {
-        static vectorN_t armatures;
+        static vectorN_t armatures_;
```
duburcqa: this is a mistake.

core/src/robot/AbstractTransmission.cc
```diff
@@ -36,42 +35,56 @@ namespace jiminy
...
+
         // Populate jointPositionIndices_
         std::vector<int32_t> jointPositionIndices;
-        returnCode = hresult_t::SUCCESS;
+        hresult_t returnCode = hresult_t::SUCCESS;
         for (std::string const & jointName : jointNames_)
         {
             std::vector<int32_t> jointPositionIdx;
-            if (!robot->model.existJointName(jointName))
+            if (!robot_->model.existJointName(jointName))
```
duburcqa: you cannot access a weak_ptr this way.

fabinsch:
https://github.com/duburcqa/jiminy/pull/432#discussion_r745625397

```
auto robot = robot_.lock();
        for (std::string const & jointName : jointNames_)
        {
            std::vector<int32_t> jointPositionIdx;
            if (!robot->model.existJointName(jointName))
```

I fixed the way how to access the robot, but it does not have a member model. I'm a bit lost how to get access to the model at this point..

duburcqa:
it is `pncModel_` not `model`. Similar to python.

fabinsch:
I still have sharedHolders all over the place in the AbstractTransmission. We agreed that we do not need this for them right ? So I will get rid of them

duburcqa:
yes, there is no shared memory for them. since nobody wants to access the state of all the transmissions at once. It is only the case for the motors and sensors, and system state.

fabinsch:
Do we still need Options for the transmission ? To set for example to mechanical reduction ratio in case of a simple transmission ?

duburcqa:
Yes, not to mention custom, more advanced transmissions for which is it hard to predict what the parameters could be. Why not some backlash parameters for instance.

fabinsch:
Ok, I'm trying to understand it using the `AbstractMotor`, and I'm wonderig why do we have `baseMotorOptions_` and `motorOptionsHolder_`

one is Structure with the parameters of the motor and public and the other Dictionary with the parameters of the motor protected

duburcqa:
> one is Structure with the parameters of the motor and public and the other Dictionary with the parameters of the motor protected

Dictionary are convenient but manipulating them is painfully slow because they are weakly typed (like python), so you need to pay the cost to infer the type at runtime. On the contrary, structure are strongly typed and thus very fast but not generic. By giving access to the users to the structure, one can make sure the end-user is accessing the options through the right API. 

In the meantime, the user is expected to get/set the options using a dictionary for convenience. (Nobody expects options to be strongly typed for the getter and setter in my opinion). The dictionary provided by the user is stored internally as a dict and returned as const by the getter. This way, it avoids having to implement a method to convert structure into dictionary for this, only the other way around is implemented. 

To sum up. Dict is used for getter and setter, since taking time for this is not a big deal. It is only done seldomly. Then user is expected to access the values at runtime using the structure, which is crazy fast. It is the only way to get convenient setter/getter interface, while runtime access as fast as possible. It is true that storing the dict could be avoided by implementing a converter from struct to dict, but I think it is better to avoid it by storing the dict.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -210,12 +210,14 @@ namespace jiminy
...
-        ///             input arguments.
+        ///             input arguments. This transformation is used to compute the forward transformation
+        ///             of the transmission transforming motor position, velocity and torque to joint
+        ///             position, velocity and torque.
         ///
         /// \param[in]  q        Current configuration of the motors.
         /// \param[in]  v        Current velocity of the motors.
         ///////////////////////////////////////////////////////////////////////////////////////////////
-        virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & q,
+        virtual float64_t computeTransform(Eigen::VectorBlock<vectorN_t const> const & q,
```
duburcqa: should be void.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -227,15 +229,15 @@ namespace jiminy
         /// \param[in]  q        Current configuration of the motors.
         /// \param[in]  v        Current velocity of the motors.
         ///////////////////////////////////////////////////////////////////////////////////////////////
-        virtual void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & q,
+        virtual float64_t computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & q,
```
duburcqa: should be void.

core/include/jiminy/core/robot/Robot.h
```diff
@@ -160,6 +161,7 @@ namespace jiminy
         std::unordered_map<std::string, bool_t> sensorTelemetryOptions_;
         std::vector<std::string> motorsNames_;                                      ///< Name of the motors
         std::unordered_map<std::string, std::vector<std::string> > sensorsNames_;   ///< Name of the sensors
+        std::vector<std::string> transmissionsNames_;                                      ///< Name of the motors
```
duburcqa: spacing `       ///<`

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -44,11 +44,11 @@ namespace jiminy
         virtual hresult_t setOptions(configHolder_t const & transmissionOptions) final override;
 
     private:
-        virtual float64_t computeTransform(Eigen::VectorBlock<vectorN_t const> /* q */,
-                                           Eigen::VectorBlock<vectorN_t const> /* v */) final override;
+        void computeTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/ ,
```
duburcqa: It should be virtual and final.

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -44,11 +44,11 @@ namespace jiminy
...
+                              Eigen::VectorBlock<vectorN_t const> const & /*v*/) override;
 
-        virtual float64_t computeInverseTransform(Eigen::VectorBlock<vectorN_t const> /* q */,
-                                                  Eigen::VectorBlock<vectorN_t const> /* v */) final override;
+        void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & /* q */,
```
duburcqa: same.

core/src/robot/BasicTransmissions.cc
```diff
@@ -17,26 +17,23 @@ namespace jiminy
         setOptions(getOptions());
     }
 
-    float64_t SimpleTransmission::computeTransform(Eigen::VectorBlock<vectorN_t> /* q */,
+    void SimpleTransmission::computeTransform(Eigen::VectorBlock<vectorN_t> /* q */,
                                                    Eigen::VectorBlock<vectorN_t> /* v */))
```
duburcqa: spacing.

core/src/robot/BasicTransmissions.cc
```diff
@@ -17,26 +17,23 @@ namespace jiminy
...
             PRINT_ERROR("Transmission not initialized. Impossible to compute actual transmission effort.");
             return hresult_t::ERROR_INIT_FAILED;
         }
-
-        return transmissionOptions_->mechanicalReduction;
     }
 
-    float64_t SimpleTransmission::computeInverseTransform(Eigen::VectorBlock<vectorN_t> /* q */,
+    void SimpleTransmission::computeInverseTransform(Eigen::VectorBlock<vectorN_t> /* q */,
                                                           Eigen::VectorBlock<vectorN_t> /* v */))
```
duburcqa: spacing.

python/jiminy_pywrap/src/Motors.cc
```diff
@@ -26,22 +26,22 @@ namespace python
         void visit(PyClass & cl) const
         {
             cl
-                .add_property("is_initialized", bp::make_function(&AbstractMotorBase::getIsInitialized,
-                                                bp::return_value_policy<bp::copy_const_reference>()))
+                // .add_property("is_initialized", bp::make_function(&AbstractMotorBase::getIsInitialized,
```
duburcqa: uncomment this one. It still makes sense.

python/jiminy_pywrap/src/Motors.cc
```diff
@@ -93,9 +93,9 @@ namespace python
 
             static void visit(PyClass & cl)
             {
-                cl
-                    .def("initialize", &TMotor::initialize)
-                    ;
+                // cl
```
duburcqa: same.

python/jiminy_pywrap/src/Transmissions.cc
```diff
@@ -0,0 +1,125 @@
...
+                .add_property("joint_idx", bp::make_function(&AbstractTransmissionBase::getJointModelIdx,
+                                            bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_type", bp::make_function(&AbstractTransmissionBase::getJointType,
+                                            bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_position_idx", bp::make_function(&AbstractTransmissionBase::getJointPositionIdx,
+                                                    bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_velocity_idx", bp::make_function(&AbstractTransmissionBase::getJointVelocityIdx,
+                                                    bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("command_limit", bp::make_function(&AbstractTransmissionBase::getCommandLimit,
+                                              bp::return_value_policy<bp::copy_const_reference>()))
```
duburcqa: spacing.

python/jiminy_pywrap/src/Transmissions.cc
```diff
@@ -0,0 +1,125 @@
...
+                                            bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_idx", bp::make_function(&AbstractTransmissionBase::getJointModelIdx,
+                                            bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_type", bp::make_function(&AbstractTransmissionBase::getJointType,
+                                            bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_position_idx", bp::make_function(&AbstractTransmissionBase::getJointPositionIdx,
+                                                    bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_velocity_idx", bp::make_function(&AbstractTransmissionBase::getJointVelocityIdx,
+                                                    bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("command_limit", bp::make_function(&AbstractTransmissionBase::getCommandLimit,
```
duburcqa: "command_limit" is a property of the motors, not the transmission !

python/jiminy_pywrap/src/Transmissions.cc
```diff
@@ -0,0 +1,125 @@
...
+                                              bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("armature", bp::make_function(&AbstractTransmissionBase::getArmature,
```
duburcqa: same for armatures ! it is for motors.

python/jiminy_pywrap/src/Transmissions.cc
```diff
@@ -0,0 +1,125 @@
...
+        ///////////////////////////////////////////////////////////////////////////////
+        /// \brief Expose.
+        ///////////////////////////////////////////////////////////////////////////////
+        static void expose()
+        {
+            bp::class_<SimpleTransmission, bp::bases<AbstractTransmissionBase>,
+                       std::shared_ptr<SimpleTransmission>,
+                       boost::noncopyable>("SimpleTransmission",
+                       bp::init<std::string const &>(
+                       bp::args("self", "Transmission_name")))
```
duburcqa: no uppercase: `transmission_name`

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        /// \details    It assumes that the internal state of the robot is consistent with the
+        ///             input arguments. This transformation is used to compute the forward transformation
+        ///             of the transmission transforming motor position, velocity and torque to joint
+        ///             position, velocity and torque.
+        ///
+        /// \param[in]  q        Current configuration of the motors.
+        /// \param[in]  v        Current velocity of the motors.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & q,
+                                      Eigen::VectorBlock<vectorN_t const> const & v) = 0;  /* copy on purpose */
```
duburcqa: It should rather be:
```
        virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & q,
                                      Eigen::VectorBlock<vectorN_t const> const & v,
                                      matrixN_t & out) = 0;  /* copy on purpose */
```
To make sure the user is setting the right matrix.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        /// \brief      Request the transmission to update its actual state based of the input data.
+        ///
+        /// \details    It assumes that the internal state of the robot is consistent with the
+        ///             input arguments.
+        ///
+        /// \param[in]  q        Current configuration of the motors.
+        /// \param[in]  v        Current velocity of the motors.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & q,
+                                             Eigen::VectorBlock<vectorN_t const> const & v) = 0;  /* copy on purpose */
```
duburcqa: same. Add out.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Compute backward transmission.
+        ///
+        /// \details    Compute backward transmission from joint to motor.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual hresult_t computeBackward(float64_t const & t,
+                                  vectorN_t const & q,
+                                  vectorN_t const & v,
+                                  vectorN_t const & a,
+                                  vectorN_t const & uJoint) final;
```
duburcqa: Add these getters:
```                                  
        matrixN_t const & getForwardTransform(void) const;
        matrixN_t const & getInverseTransform(void) const;
```

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        /// \brief      Get the configuration options of the transmission.
+        ///
+        /// \return     Dictionary with the parameters of the transmission
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        configHolder_t getOptions(void) const;
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Get the actual state of the transmission at the current time.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        float64_t const & get(void) const;
```
duburcqa: Remove this method. It does not make sense here.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        /// \details    It is the index of the joints associated with the transmission in the velocity vector.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        vectorN_t const & getJointVelocityIndices(void) const;
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Get motorName_.
+        ///
+        /// \details    It is the name of the motors associated with the transmission.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        std::vector<std::string> const & getMotorNames(void) const;
```
duburcqa, line 179: Add `getMotorIndices`

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        std::vector<std::string> jointNames_;
+        std::vector<jointIndex_t> jointModelIndices_;
+        std::vector<joint_t> jointTypes_;
+        vectorN_t jointPositionIndices_;
+        vectorN_t jointVelocityIndices_;
+        std::vector<std::string> motorNames_;
+        std::vector<std::weak_ptr<AbstractMotorBase> > motors_;
+        matrixN_t forwardTransform_;
+        matrixN_t backwardTransform_;
+
```
duburcqa: remove space.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        bool_t isAttached_;                                          ///< Flag to determine whether the transmission is attached to a robot
+        std::weak_ptr<Robot const> robot_;                           ///< Robot for which the command and internal dynamics
+        std::string name_;                                           ///< Name of the transmission
+        int32_t transmissionIdx_;                                           ///< Index of the transmission in the transmission buffer
+        std::vector<std::string> jointNames_;
+        std::vector<jointIndex_t> jointModelIndices_;
+        std::vector<joint_t> jointTypes_;
+        vectorN_t jointPositionIndices_;
+        vectorN_t jointVelocityIndices_;
+        std::vector<std::string> motorNames_;
```
duburcqa, line 289: store motors indices in addition.

core/include/jiminy/core/robot/BasicMotors.h
```diff
@@ -56,16 +56,10 @@ namespace jiminy
         auto shared_from_this() { return shared_from(this); }
         auto shared_from_this() const { return shared_from(this); }
 
-        hresult_t initialize(std::string const & jointName);
```
duburcqa, line 59: You still need an init method. Just the signature is simpler:
```
hresult_t initialize(void);
```

core/include/jiminy/core/robot/Model.h
```diff
@@ -26,14 +27,16 @@ namespace jiminy
...
     };
 
-    std::array<constraintsHolderType_t, 4> const constraintsHolderTypeRange = {{
+    std::array<constraintsHolderType_t, 5> const constraintsHolderTypeRange = {{
         constraintsHolderType_t::BOUNDS_JOINTS,
         constraintsHolderType_t::CONTACT_FRAMES,
         constraintsHolderType_t::COLLISION_BODIES,
-        constraintsHolderType_t::USER
+        constraintsHolderType_t::USER,
+        constraintsHolderType_t::TRANSMISSIONS
```
duburcqa, line 39: Move this in another PR.

core/include/jiminy/core/robot/Model.h
```diff
@@ -91,6 +94,9 @@ namespace jiminy
                 case constraintsHolderType_t::CONTACT_FRAMES:
                     constraintsMapPtr = &contactFrames;
                     break;
+                case constraintsHolderType_t::TRANSMISSIONS:
```
duburcqa, line 97: same.

core/include/jiminy/core/robot/Model.h
```diff
@@ -115,6 +121,7 @@ namespace jiminy
     public:
         constraintsMap_t boundJoints;                   ///< Store internal constraints related to joint bounds
         constraintsMap_t contactFrames;                 ///< Store internal constraints related to contact frames
+        constraintsMap_t transmissionConstraints;       ///< Store internal constraints related to the transmissions
```
duburcqa, line 124: same.

core/include/jiminy/core/robot/Robot.h
```diff
@@ -72,6 +80,7 @@ namespace jiminy
                                   vectorN_t const & command);
         vectorN_t const & getMotorsEfforts(void) const;
         float64_t const & getMotorEffort(std::string const & motorName) const;
+        float64_t getMotorEffortLimit(void);
```
duburcqa: Rather `vectorN_t const & getMotorsEffortLimits(void) const;`

core/src/engine/EngineMultiRobot.cc
```diff
@@ -3682,6 +3687,8 @@ namespace jiminy
         pinocchio::Model const & model = system.robot->pncModel_;
         pinocchio::Data & data = system.robot->pncData_;
 
+        // TODO wtf
+        // data.rotorInertia = system.robot->getArmature();
```
duburcqa: Unocmment and fix the typo.
```
data.rotorInertia = system.robot->getArmatures();
```

the problem here is the following erropr
> class "pinocchio::DataTpl<double, 0, pinocchio::JointCollectionDefaultTpl>" has no member "rotorInertia"C/C++(135)

core/src/robot/AbstractMotor.cc
```diff
@@ -216,15 +231,6 @@ namespace jiminy
             }
         }
 
-        if (returnCode == hresult_t::SUCCESS)
```
duburcqa: Not not remove this block. It still makes sense.

core/src/robot/AbstractTransmission.cc
```diff
@@ -0,0 +1,358 @@
...
+            }
+            jointIndex_t const & jointModelIdx = robot->pncModel_.getJointId(jointName);
+            int32_t const & jointVelocityFirstIdx = robot->pncModel_.joints[jointModelIdx].idx_v();
+            int32_t const & jointNv = robot->pncModel_.joints[jointModelIdx].nv();
+            jointVelocityIdx.resize(jointNv);
+            std::iota(jointVelocityIdx.begin(), jointVelocityIdx.end(), jointVelocityFirstIdx);
+            jointVelocityIndices.insert(jointVelocityIndices.end(), jointVelocityIdx.begin(), jointVelocityIdx.end());
+        }
+
+        // missing how to populate vectorN_t jointVelocityIndices_
```
duburcqa: Add TODO keyword.

core/src/robot/AbstractTransmission.cc
```diff
@@ -0,0 +1,358 @@
...
+        }
+
+        // Make sure the robot still exists
+        if (robot.expired())
+        {
+            PRINT_ERROR("Robot pointer expired or unset.");
+            return hresult_t::ERROR_GENERIC;
+        }
+
+        // Make sure the joint is not already attached to a transmission
```
duburcqa: Add TODO keyword.

core/src/robot/BasicMotors.cc
```diff
@@ -17,23 +17,6 @@ namespace jiminy
         setOptions(getDefaultMotorOptions());
     }
 
-    hresult_t SimpleMotor::initialize(std::string const & jointName)
```
duburcqa, line 20: Do not remove this whole method. Just the part related to joint name.

core/src/robot/Model.cc
```diff
@@ -46,6 +46,7 @@ namespace jiminy
     constraintsHolder_t::constraintsHolder_t(void) :
     boundJoints(),
     contactFrames(),
+    transmissionConstraints(),
```
duburcqa, line 49: Other PR.

core/src/robot/Model.cc
```diff
@@ -101,6 +102,7 @@ namespace jiminy
                 break;
             case constraintsHolderType_t::USER:
             case constraintsHolderType_t::COLLISION_BODIES:
+            case constraintsHolderType_t::TRANSMISSIONS:
```
duburcqa, line 105: same.

core/src/robot/Model.cc
```diff
@@ -171,6 +173,7 @@ namespace jiminy
             collisionBodies.push_back(constraintsMap);
             break;
         case constraintsHolderType_t::USER:
+        case constraintsHolderType_t::TRANSMISSIONS:
```
duburcqa, line 176: same.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+        /// \details    It is the name of the motors associated with the transmission.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        std::vector<std::string> const & getMotorNames(void) const;
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Compute forward transmission.
+        ///
+        /// \details    Compute forward transmission from motor to joint.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual hresult_t computeForward(float64_t const & t,
```
duburcqa: It should not be virtual. Remove also final keyword.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,278 @@
...
+                                 vectorN_t & v,
+                                 vectorN_t & a,
+                                 vectorN_t & uJoint) final;
+
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        /// \brief      Compute backward transmission.
+        ///
+        /// \details    Compute backward transmission from joint to motor.
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual hresult_t computeBackward(float64_t const & t,
```
duburcqa: same.

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -0,0 +1,58 @@
...
+
+        virtual hresult_t setOptions(configHolder_t const & transmissionOptions) final override;
+
+    private:
+        virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/ ,
+                                      Eigen::VectorBlock<vectorN_t const> const & /*v*/) final override;
+
+        virtual void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/,
+                                             Eigen::VectorBlock<vectorN_t const> const & /*v*/) final override;
+
```
duburcqa, line 56: You forgot to overload `virtual void computeEffortTransmission(void)`

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -218,7 +234,8 @@ namespace jiminy
         /// \param[in]  v        Current velocity of the motors.
         ///////////////////////////////////////////////////////////////////////////////////////////////
         virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & q,
-                                      Eigen::VectorBlock<vectorN_t const> const & v) = 0;  /* copy on purpose */
+                                      Eigen::VectorBlock<vectorN_t const> const & v,
+                                      matrixN_t & out) = 0;  /* copy on purpose */
```
duburcqa: wtf there is no copy here. remove this comment  x)

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -230,8 +247,8 @@ namespace jiminy
         /// \param[in]  v        Current velocity of the motors.
         ///////////////////////////////////////////////////////////////////////////////////////////////
         virtual void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & q,
-                                             Eigen::VectorBlock<vectorN_t const> const & v) = 0;  /* copy on purpose */
-
+                                             Eigen::VectorBlock<vectorN_t const> const & v,
+                                             matrixN_t & out) = 0;  /* copy on purpose */
```
duburcqa: same.

core/include/jiminy/core/robot/BasicMotors.h
```diff
@@ -56,16 +56,11 @@ namespace jiminy
...
+        hresult_t initialize(void);
         virtual hresult_t setOptions(configHolder_t const & motorOptions) final override;
 
     private:
-        virtual hresult_t computeEffort(float64_t const & t,
-                                        Eigen::VectorBlock<vectorN_t const> const & q,
-                                        float64_t const & v,
-                                        float64_t const & a,
-                                        float64_t command) final override;
+        virtual hresult_t computeEffort(float64_t command) final override;
```
duburcqa: add comment `/* copy on purpose */`.

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -0,0 +1,62 @@
...
+        SimpleTransmission(std::string const & name);
+        virtual ~SimpleTransmission(void) = default;
+
+        auto shared_from_this() { return shared_from(this); }
+        auto shared_from_this() const { return shared_from(this); }
+
+        virtual hresult_t setOptions(configHolder_t const & transmissionOptions) final override;
+
+    private:
+        virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/,
```
duburcqa: uncomment `q` and `v`

core/include/jiminy/core/robot/BasicTransmissions.h
```diff
@@ -0,0 +1,62 @@
...
+                                      Eigen::VectorBlock<vectorN_t const> const & /*v*/,
+                                      matrixN_t & out) final override;
+
+        virtual void computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/,
```
duburcqa: same.

core/src/robot/AbstractMotor.cc
```diff
@@ -225,6 +240,7 @@ namespace jiminy
             }
         }
 
+
```
duburcqa: remove space.

core/src/robot/AbstractTransmission.cc
```diff
@@ -0,0 +1,391 @@
...
+            jointNames_.clear();
+            motorNames_.clear();
+            isInitialized_ = false;
+        }
+
+        auto robot = robot_.lock();
+
+        // TODO move this stuff to refresh Proxies ?
+        // Populate motorIndices_
+        std::weak_ptr<AbstractMotorBase const> motor;
```
duburcqa: Move this to `refreshProxies`

core/src/robot/AbstractTransmission.cc
```diff
@@ -0,0 +1,391 @@
...
+        for (std::string const & motorName : motorNames)
+        {
+            returnCode = robot->getMotor(motorName, motor);
+            auto motorTemp = motor.lock();
+            int32_t idx = motorTemp->getIdx();
+            motorIndices_.push_back(idx);
+        }
+
+        // Populate jointPositionIndices_
```
duburcqa: same.

core/src/robot/AbstractTransmission.cc
```diff
@@ -0,0 +1,391 @@
...
+            }
+        }
+        int32_t jointPositionSize = jointPositionIndices.size();
+        jointPositionIndices_.resize(jointPositionSize);
+        for (int32_t i = 0; i <  jointPositionSize; ++i)
+        {
+            jointPositionIndices_(i) = jointPositionIndices[i];
+        }
+
+        // Populate jointVelocityIndices_
```
duburcqa: same.

core/src/robot/BasicMotors.cc
```diff
@@ -17,17 +17,14 @@ namespace jiminy
...
+    hresult_t initialize(void)
```
duburcqa: missing `SimpleMotor::`

core/src/robot/BasicTransmissions.cc
```diff
@@ -0,0 +1,44 @@
...
+        setOptions(getOptions());
+    }
+
+    void SimpleTransmission::computeTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/,
+                                              Eigen::VectorBlock<vectorN_t const> const & /*v*/,
+                                              matrixN_t &out)
+    {
+        if (!isInitialized_)
+        {
+            PRINT_ERROR("Transmission not initialized. Impossible to compute actual transmission effort.");
```
duburcqa: wrong error message.

core/src/robot/BasicTransmissions.cc
```diff
@@ -0,0 +1,44 @@
...
+        }
+    }
+
+    void SimpleTransmission::computeInverseTransform(Eigen::VectorBlock<vectorN_t const> const & /*q*/,
+                                                     Eigen::VectorBlock<vectorN_t const> const & /*v*/,
+                                                     matrixN_t &out)
+    {
+        if (!isInitialized_)
+        {
+            PRINT_ERROR("Transmission not initialized. Impossible to compute actual transmission effort.");
```
duburcqa: same.

python/jiminy_pywrap/src/Motors.cc
```diff
@@ -32,16 +32,16 @@ namespace python
...
-                                            bp::return_value_policy<bp::copy_const_reference>()))
-                .add_property("joint_idx", bp::make_function(&AbstractMotorBase::getJointModelIdx,
-                                            bp::return_value_policy<bp::copy_const_reference>()))
-                .add_property("joint_type", bp::make_function(&AbstractMotorBase::getJointType,
-                                            bp::return_value_policy<bp::copy_const_reference>()))
-                .add_property("joint_position_idx", bp::make_function(&AbstractMotorBase::getJointPositionIdx,
-                                                    bp::return_value_policy<bp::copy_const_reference>()))
-                .add_property("joint_velocity_idx", bp::make_function(&AbstractMotorBase::getJointVelocityIdx,
-                                                    bp::return_value_policy<bp::copy_const_reference>()))
+                // .add_property("joint_name", bp::make_function(&AbstractMotorBase::getJointName,
```
duburcqa: remove rather than comment.

python/jiminy_pywrap/src/Transmissions.cc
```diff
@@ -0,0 +1,123 @@
...
+                .add_property("joint_types", bp::make_function(&AbstractTransmissionBase::getJointTypes,
+                                             bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_position_indices", bp::make_function(&AbstractTransmissionBase::getJointPositionIndices,
+                                                        bp::return_value_policy<bp::copy_const_reference>()))
+                .add_property("joint_velocity_indices", bp::make_function(&AbstractTransmissionBase::getJointVelocityIndices,
+                                                        bp::return_value_policy<bp::copy_const_reference>()))
+
+                .def("set_options", &PyAbstractTransmissionVisitor::setOptions)
+                .def("get_options", &AbstractTransmissionBase::getOptions)
+                .def("compute_transform", &AbstractTransmissionBase::computeTransform)
```
duburcqa, line 52: Missing getters

```
.add_property("transform", bp::make_function(&AbstractConstraintBase::getForward,
                                           bp::return_value_policy<result_converter<false> >()))
```
same for backward.

fabinsch:
https://github.com/duburcqa/jiminy/pull/432#discussion_r770577643

maybe you can explain me quickly what do you mean exactly with this "copy on purpose". I tried to do it in a same way then AbstractMotor.h
`virtual hresult_t computeEffort(float64_t command) = 0;  /* copy on purpose */`

fabinsch:
https://github.com/duburcqa/jiminy/pull/432#discussion_r770589573

why in this case you propose to do
`bp::return_value_policy<result_converter<false> >()))`

whereas the other getters have 
`bp::return_value_policy<bp::copy_const_reference>()))`

fabinsch:
https://github.com/duburcqa/jiminy/pull/432#discussion_r770581950

why you think we shouold q and v in this case? I thought the simple transmission is just a basic reduction ratio that is state independent

duburcqa:
> maybe you can explain me quickly what do you mean exactly with this "copy on purpose". 

The point is to write "copy on purpose" when you do a copy on purpose. In your example `computeEffort(float64_t command)` indeed you do a copy, on purpose by definition because by default you should do `computeEffort(float64_t const & command)`. So the point is to mention when a copy is made on purpose to not remove it afterward during review or whatever thinking it is a mistake.

duburcqa:
You can have a look at the definition of `result_converter`. It is used to decide how to handle eigen arrays returned by C++ to Python. If it is false, it means by reference, if it is true then it is by copy. For other objects other return policies must be used. For instance `bp::copy_const_reference` is to copy const reference that are return by C++ to Python (if it is directly convertible, otherwise it would fail.)
```
bp::return_value_policy<result_converter<false> >()))
```

duburcqa:
> why you think we shouold q and v in this case? I thought the simple transmission is just a basic reduction ratio that is state independent

Yes indeed, but you never comment variable names in declaration. Only in definitions.

fabinsch:
```bash
/home/fabian.schramm/wdc_workspace/src/jiminy/core/src/engine/EngineMultiRobot.cc: In member function ‘const vectorN_t& jiminy::EngineMultiRobot::computeAcceleration(jiminy::systemHolder_t&, jiminy::syst
emDataHolder_t&, const vectorN_t&, const vectorN_t&, const vectorN_t&, jiminy::forceVector_t&)’:
/home/fabian.schramm/wdc_workspace/src/jiminy/core/src/engine/EngineMultiRobot.cc:3690:14: error: ‘pinocchio::Data’ {aka ‘struct pinocchio::DataTpl<double>’} has no member named ‘rotorInertia’
 3690 |         data.rotorInertia = system.robot->getArmatures();
      |              ^~~~~~~~~~~~
```

I think this is quite werid as I did not modify something related to pinocchio:Data (that's why I had a comment with wtf before)

duburcqa:
It is expected. it should be `model` not `data`.

core/src/engine/EngineMultiRobot.cc
```diff
@@ -3682,6 +3687,7 @@ namespace jiminy
         pinocchio::Model const & model = system.robot->pncModel_;
         pinocchio::Data & data = system.robot->pncData_;
 
+        data.rotorInertia = system.robot->getArmatures();
```
duburcqa: `model`, not `data`.

duburcqa:
I was think, and there is still an issue with the API of the transmission. It is not possible to define a transmission only by its transform matrix (jacobian) because it only applies on spatial motion and force vectors (velocity and upward) but not on the configuration. For this, you should provided another methods, converting back and forth configuration vectors from joint space to motor space. There is no other choice to be generic.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,297 @@
...
+        /// \details    It assumes that the internal state of the robot is consistent with the
+        ///             input arguments. This transformation is used to compute the forward transformation
+        ///             of the transmission transforming motor position, velocity and torque to joint
+        ///             position, velocity and torque.
+        ///
+        /// \param[in]  q        Current configuration of the motors.
+        /// \param[in]  v        Current velocity of the motors.
+        /// \param[out] out      Matrix transforming quantities from motor to joint level
+        ///////////////////////////////////////////////////////////////////////////////////////////////
+        virtual void computeTransform(Eigen::VectorBlock<vectorN_t const> const & q,
```
duburcqa, line 237: Rename them in `computeJacobian` and `computeInverseJacobian` instead. 

And add new methods `computeTransform` and `computeInverseTransform`, which works at position level only. Make them public because the user may want to call them manually.
```
        virtual void computeTransform(vectorN_t const & qMotors,
                                                          vectorN_t          & qJoints) = 0;
```

duburcqa, line 237: Remove `v` dependency. We will stick to holonomic transmission. It is much simpler and cover 99% of real applications.

duburcqa, line 237: Change `Eigen::VectorBlock<vectorN_t const>` for `vectorN_t const`.

duburcqa, line 237: Explicit what is `q`. `qMotors` or `qJoints`

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,297 @@
...
+                                      Eigen::VectorBlock<vectorN_t const> const & v,
+                                      matrixN_t & out) = 0;
```
duburcqa, line 239: Rename `out` in `Jac` for clarity.

core/include/jiminy/core/robot/AbstractTransmission.h
```diff
@@ -0,0 +1,297 @@
...
+        int32_t transmissionIdx_;                                           ///< Index of the transmission in the transmission buffer
+        std::vector<std::string> jointNames_;
+        std::vector<jointIndex_t> jointModelIndices_;
+        std::vector<joint_t> jointTypes_;
+        vectorN_t jointPositionIndices_;
+        vectorN_t jointVelocityIndices_;
+        std::vector<std::string> motorNames_;
+        std::vector<int32_t> motorIndices_;
+        std::vector<std::weak_ptr<AbstractMotorBase> > motors_;
+        matrixN_t forwardTransform_;
```
duburcqa, line 292: `forwardJacobian_`

duburcqa, line 292: Add public getter for `forwardJacobian_` and `backwardTransform_`.

duburcqa:
After doing the math, I realized you should remove the acceleration from motor state, and from transmission forward/backward. It is more tricky to handle acceleration (it requires any additional user specified matrix) and not necessary in practice in 99% of the usecases.

duburcqa:
You should also limit yourself to invertible transmissions,. Which are n-to-n motor to joint mappings. And rename the abstract case accordingly. Non-invertible transmission must be modelled through constraints instead.

duburcqa:
It is necessary to access the motors directly at robot level for several reasons, including the calling the *All methods of motors (such as resetAll ), which cannot be done at transmission level since there is no sharedHolder for them. So I suggest still dealing with the motor directly at robot level, instead of doing it through the transmission as we suggested. Similarly, I tend to think it may be better to remove again notifyRobot, and just create and return the list when the getter is called (no longer returning an attribute but a temporary). For me using notifyRobot is too much of a hassle just to provide a convenience getter. Also, you should add a reset method to the transmission, and reset them alongside the motors.

duburcqa:
Cassie could be the first robot using an advanced transmission in Jiminy. Currently it is done by solving a optimization problem, which is likely to be more costly. Could be interesting to compare both. Transmission should be unbeatable. Ideally we should also implement joint spring. So we could compare the "exact" model with the simplified one with linearized transmission.

duburcqa:
`rotorInertia` should be a config option of the motors, and the transmission inertia should be accessed by the robot through the transmissions. This is necessary for advanced transmission. Ideally, once simple transmission are implemented, it would be great to add exact rotor inertia modelling instead of relying on the classical approximation by diagonal terms on the mass matrix.

PR #457, opened 2021-11-29
[core] Fix PGS cone bounds.

duburcqa:


duburcqa:
It is not numerical stable for some reason. Simulation of Atalante robot explodes during training.

PR #459, opened 2021-11-30
[core] Compute PGS error in residual space.

duburcqa:
It helps PGS solver to "converge" in case of redundant constraints since the residuals are always supposed to converge contrary to the optimization variables.


duburcqa:
For some reason the gym integration unit test are no longer passing. Probably some numerical instability for some reason. Closing for now.

PR #478, opened 2022-01-12
Fixes some field in model not being updated when adding flexibilities.

matthieuvigne:
When adding a flexibility in the model, the joints are 'manually' reordered. pinocchio 2.0 added some new members to the Model class, these members were not update and where thus incorrect. 

This only impacted hessian computation - but would cause a crash when compiled in Debug.


matthieuvigne:
Closing as already fixed in duburcqa:master

duburcqa:
Ah interesting ! I was not aware of the impact, just that now the CI is running in release and debug mode, so I caught the bug and fixed it. Many bugs have been fixed on master wrt wandercraft pinned version, and even some more on dev branch !

PR #479, opened 2022-01-20
[python/viewer] Fallback to generic direct connection through ipykernel for Meshcat.

duburcqa:
Several limitations:
* Getting the server origin is difficult. `window.location.origin` is not reliable, for instance it will fail if the notebook server is hidden behind a reverse proxy, or it is embedded in a app such as VSCode. Using a tunnel and go through the usual meshcat zmq socket is an alternative solution, but it requires setting up a server or paying a service for it. Another option would be to develop a proper widget to get access to the kernel websocket through it, but it is not solution the compatibility issue with cloud notebooks providers such as deepnote, kaggle or colab.
* It is suffering from throughput limitation that are monitored by ipython kernel. The solution would be to gather the messages whenever it is possible, for instance when updating the position for frames.
* Managing the websocket in the iFrame is problematic since when the window unload it gets closed and the delay is too short to notify the kernel in order to keep track of the number of communication opened.

To sum up, finding a way to get access to the kernel websocket instead of opening a communication manually inside the frame seems to be necessary, but there is no way to do so by design... So being generic looks like a dead end. But the current implementation is good enough to more or less work on several platforms such as mybinder and kaggle (with a little tweak of the url origin).

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -47,7 +47,7 @@ jobs:
 
         "${PYTHON_EXECUTABLE}" -m pip install --upgrade pip
         "${PYTHON_EXECUTABLE}" -m pip install --upgrade wheel
-        "${PYTHON_EXECUTABLE}" -m pip install --upgrade numpy
+        "${PYTHON_EXECUTABLE}" -m pip install --upgrade "numpy<1.22"  # for numba compat.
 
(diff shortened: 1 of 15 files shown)


duburcqa:
It also fixes `has_gui` in interactive mode.

PR #533, opened 2022-06-08
[python/viewer] Load collada mesh with texture in Meshcat.

duburcqa:


python/jiminy_py/src/jiminy_py/viewer/meshcat/index.html
```diff
@@ -14,15 +14,46 @@
...
+                        img.onerror = reject
+                        img.src = url;
+                    })
+                }
+                return preloaded_resources[key];
+            }
+
             var handle_command = viewer.handle_command;
-            viewer.handle_command = function(cmd) {
+            viewer.handle_command = async function(cmd) {
```
jcarpent, line 44: What is the purpose of this command?

python/jiminy_py/src/jiminy_py/viewer/meshcat/meshcat_visualizer.py
```diff
@@ -37,75 +40,124 @@ def lower(self, object_data: Any) -> MsgType:
...
+                                (i + 1) * N + j])
+                faces.append(vec[[0, 1, 2]])
+                faces.append(vec[[0, 2, 3]])
+        faces = np.vstack(faces)
+
+        # Init base class
+        super().__init__(vertices, faces)
+
+
+class DaeMeshGeometryWithTexture(ReferenceSceneElement):
```
jcarpent, line 86: Could you port it back to MeshCat?
I think it can be helpful to many people beyond this project. Don't you think so?

duburcqa: Yes, there is already a PR opened on the official meshcat-python repository https://github.com/rdeits/meshcat-python/pull/112. I'm getting involved but not contributing directly. I'm afraid it is going to take quite some time, mainly because the owner is quite reluctant to merge new contributions. It was more of a side project just to play around for a few hours and I'm not sure it is robust enough to be integrated in meshcat itself as is. I'm even considering dropping meshcat completely.

python/jiminy_py/src/jiminy_py/viewer/meshcat/index.html
duburcqa: You mean the patch of the original command ? It is necessary to wait for the required textures to finish loading before processing the original `set_object` command loading the geometry with materials in threejs. This design enables to send and load shared textures only once. It speeds up the initialization of the threejs view up to a factor 5 for robots like Atlas.

Besides, I'm also patching it to handle  additional commands including bi-directional communication.

PR #569, opened 2022-09-12
[misc] Fix required versions of some gym-jiminy dependencies.

nicolasWDC:
add and correct dependency

### DIFF
--- a/python/gym_jiminy/common/setup.py
+++ b/python/gym_jiminy/common/setup.py
@@ -62,7 +62,8 @@ setup(
         # - 0.18.0: dtype handling of flatten space
         # - >=0.18.0,<0.18.3 requires Pillow<8.0 to work, not compatible with
         #   Python 3.9.
-        "gym>=0.18.3"
+        "gym>=0.18.3,<0.21.0"
+        "scipy<=1.7.3"
     ],
     extras_require=extras,
     zip_safe=False


python/gym_jiminy/common/setup.py
```diff
@@ -62,7 +62,8 @@
         # - 0.18.0: dtype handling of flatten space
         # - >=0.18.0,<0.18.3 requires Pillow<8.0 to work, not compatible with
         #   Python 3.9.
-        "gym>=0.18.3"
+        "gym>=0.18.3,<0.21.0"
+        "scipy<=1.7.3"
```
duburcqa, line 66: `scipy` is actually a dependency of [gym-jiminy-toolbox](https://github.com/duburcqa/jiminy/blob/686977c798234cf4ed87907a76ca670798e1c36d/python/gym_jiminy/toolbox/setup.py#L40) and `1.7.3` is already the latest version before `1.8.0`, so it should be picked anyway.

python/gym_jiminy/common/setup.py
```diff
@@ -62,7 +62,8 @@
         # - 0.18.0: dtype handling of flatten space
         # - >=0.18.0,<0.18.3 requires Pillow<8.0 to work, not compatible with
         #   Python 3.9.
-        "gym>=0.18.3"
+        "gym>=0.18.3,<0.21.0"
```
duburcqa, line 65: `gym>=0.18.3,<=0.21.0` is actually supported.

python/gym_jiminy/common/setup.py
```diff
@@ -62,7 +62,8 @@
         # - 0.18.0: dtype handling of flatten space
         # - >=0.18.0,<0.18.3 requires Pillow<8.0 to work, not compatible with
         #   Python 3.9.
```
duburcqa, line 64:
```suggestion
        #   Python 3.9.
        #   - 0.22.0: major public API breakage. 
```

python/gym_jiminy/common/setup.py
duburcqa: By the way, I think it is only [gym-jiminy-rllib](https://github.com/duburcqa/jiminy/blob/3432d600e88d921378f6c34af131d1f7ac0a7dcf/python/gym_jiminy/rllib/setup.py#L32) that actually requires `<=0.21.0` (instead of `<=0.24.0` for `gym_jiminy-common` on dev branch). I have not double checked though.

Issue #337, opened 2021-05-01
[python/viewer] Update contact sensor data when replaying log file.


duburcqa:
It only requires to restore sensors and motors originally attached to the robot when building robot from log, which is not the case for now. This could be done by enabling exporting hardware configuration file from a robot.

duburcqa:
Similarly, the theoretical model is lost, which means that it is theoretically impossible to load the options from the configuration file, which is problematic. The good procedure would be to restore first the theoretical model, then to update the options (and the hardware), and finally to overwrite the theoretical model with the actual one.

duburcqa:
Fixed by https://github.com/duburcqa/jiminy/pull/775

PR #572, opened 2022-09-25
[core|python] Unify log reading/writing.

duburcqa:
* [core|python] Unify log reading/writing.
* [core|python] Remove support of log CSV format.


duburcqa:
The motors and sensors are not restored after loading a log file. They should be serialized along with any other information. Yet, doing so is not straightforward at the time being... https://github.com/duburcqa/jiminy/issues/337

PR #579, opened 2022-10-03
Bugfix: ConstraintBlock members are uninitialized since f61ba116.

matthieuvigne:
Commit f61ba116 accidentally removed the initialization of ConstraintBlock fields, causing weird behavior down the road. The init is added back, to the values that were present before this commit.

### DIFF
--- a/core/include/jiminy/core/solver/ConstraintSolvers.h
+++ b/core/include/jiminy/core/solver/ConstraintSolvers.h
@@ -29,8 +29,7 @@ namespace jiminy
     public:
         AbstractConstraintBase * constraint;
         Eigen::Index startIdx;
-        bool_t isBounded;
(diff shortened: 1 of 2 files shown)


duburcqa:
Nice catch thank you ! But I don't think this is the right patch. This real problem is plain struct are NOT default constructed automatically, only memory is allocated if not explicitly specified otherwise. Could you try the one I propose instead ?

duburcqa:
If you want to write a unit test, you can have a look at this [example script](https://github.com/duburcqa/jiminy/blob/dev/python/gym_jiminy/examples/box_uneven_ground_impulse_contact.py).

duburcqa:
I force-pushed on your branch. I hope this time it should be fixed.

PR #584, opened 2022-10-25
[core] Fix constraints relying on wrong acceleration (classical not spatial).

duburcqa:
* [core] Fix constraints relying on wrong acceleration (classical not spatial).
* [core] Fix sphere and whee constraints.
* [python/viewer] Fix exception when no collision bodies are specified.
* [python] Add example scripts.


core/src/constraints/WheelConstraint.cc
```diff
@@ -110,22 +110,24 @@ namespace jiminy
...
+        vector3_t const & omega = frameVelocity.angular();
         auto dx3_ = - omega.cross(x3_);  // Using auto to not evaluate the expression
         pinocchio::alphaSkew(radius_, dx3_, dskewRadius_);
 
         // Compute frame drift in local frame
-        pinocchio::Motion const driftLocal = getFrameAcceleration(model->pncModel_,
-                                                                  model->pncData_,
-                                                                  frameIdx_,
-                                                                  pinocchio::LOCAL);
+        pinocchio::Motion driftLocal = getFrameAcceleration(model->pncModel_,
```
matthieuvigne: You can directly use `getFrameClassicalAcceleration` for this result.

duburcqa: I could, but it is slower, since it calls `getFrameVelocity` internally, which is not returning a reference but doing the computations ever and ever again.

PR #595, opened 2023-03-11
Jiminy release 1.7.15

duburcqa:


.github/workflows/linux.yml
```diff
@@ -125,3 +125,29 @@ jobs:
...
+        cp $RootDir/build/stubs/jiminy_py-stubs/core/__init__.pyi \
+           $RootDir/build/pypi/jiminy_py/src/jiminy_py/core/core.pyi
+
+        # Re-install jiminy with stubs
+        cd "$RootDir/build"
+        make install
+
+        # mypy is not able to follow `.pth` when checking directories or files.
+        # This bug may be fixed in the future. The only workaround is to set `MYPYPATH` manually.
+        # Anyway, the stubs generated for eigenpy, hppfcl and pinocchio they are incomplete and
```
jcarpent, line 145: Why do claim that here?!

jcarpent: If there are errors, it would be kind to fix them and contribute to the original source you took inspiration from (https://github.com/jcarpent/pybind11-stubgen). I guess the use of open-source projects should go in both ways.

duburcqa: > Why do claim that here?!

Just to remember for later why I'm not enabling stub generation on these projects for now. Nothing more.  There is work to do upstream to improve the situation. I just don't want to do it now as it is not a big priority for me.

PR #632, opened 2023-08-15
[gym_jiminy/common] Speed-up nested data structure utilities.

duburcqa:


python/gym_jiminy/common/gym_jiminy/common/utils/spaces.py
```diff
@@ -263,21 +319,91 @@ def copy(data: DataNestedT) -> DataNestedT:
...
+        return _clip_or_copy(data, space)
+
+    out = OrderedDict()
+    for field, subspace in dict.items(space.spaces):
+        out[field] = clip(data[field], subspace)
+    return out
+
+
+def contains(data: DataNestedT, space: gym.Space[DataNestedT]) -> bool:
+    """TODO: Add documentation
```
duburcqa: Check if all leaves of a nested data structure are within bounds of there respective `gym.Space`.

By design, it is always `True` for all spaces but `gym.spaces.Box`, `gym.spaces.Discrete` and `gym.spaces.MultiDiscrete`.

duburcqa:
Move space type logics in specialized clip and contains builders without indirection to avoid check at runtime. (Plus allow skipping leaves that are not affected or not changing the results).

duburcqa:
Small gain could be expected by specializing `contains` (~1.0%).

PR #635, opened 2023-08-19
[core] Improve telemetry performance.

### LINKED ISSUE #33: [core] Improve performance of the telemetry
`telemetrySender_.registerNewEntry` should return a handle with `updateValue` already specialized to this specific variable, so that it avoids having to search it in the whole dictionary of registered variables systematically.

duburcqa:
Atlas learning environment now has a real time factor of 55 on a single core.

* [core] Skip useless forward kinematics when updating telemetry.
* [core] Refactor telemetry sender to improve efficiency.
* [gym_jiminy/common] PID block supports ordering mismatch between sensors and motors.
* [gym_jiminy/common] Remove useless action buffer from ControllerBlock.
* [gym_jiminy/common] Speed-up refresh observation.
* [gym_jiminy/common] Speed up PID controller block.
* [gym_jiminy] Do not release GIL for small helpers.

Fixes https://github.com/duburcqa/jiminy/issues/33

### DIFF
--- a/core/include/jiminy/core/control/AbstractController.h
+++ b/core/include/jiminy/core/control/AbstractController.h
@@ -76,12 +76,27 @@ namespace jiminy
 
         ///////////////////////////////////////////////////////////////////////////////////////////////
         ///
-        /// \brief      Dynamically registered a Eigen Vector to the telemetry.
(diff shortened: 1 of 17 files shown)


duburcqa:
profile

duburcqa:
profile

PR #716, opened 2024-02-12
[core] Fix error estimation too optimistic for adaptive steppers.

duburcqa:


core/src/stepper/runge_kutta_dopri_stepper.cc
```diff
@@ -36,44 +36,45 @@ namespace jiminy
...
-        if (tolRel_ > EPS)
-        {
-            otherSolution_.setZero();
-            solution.difference(otherSolution_, scale_);
-            error_ /= scale_;
-            errorRelNorm = error_.normInf() / tolRel_;
-        }
+        // Compute error scale
+        otherSolution_.setZero();
+        solution.difference(otherSolution_, scale_);
```
duburcqa: It would be more robust to take the initial state as reference, so that the scale does not depend on the integration itself.
```
initialState.difference(otherSolution_, scale_);
```

PR #742, opened 2024-03-18
[core] Fix some symbols not exported by mistake.

duburcqa:


duburcqa:
It would be worth trying to remove the explicit template instantiation and see if it works.

PR #743, opened 2024-03-18
[misc] Fix Python detection edge-cases.

duburcqa:


duburcqa:
False alarm. It was coming from a mis-configuration of custom user toolchain.

PR #745, opened 2024-03-20
[core] Avoid explicit template instantiation if possible.

duburcqa:


duburcqa:
Actually It does not work. Closing.

Issue #246, opened 2020-12-10
[python] Full support of EngineMultirobot in Python.


duburcqa:
`Simulator` should be modified to support both Engine and EngineMultirobot. The same goes for `Viewer`.

In the current implementation of Engine, some base methods are explicitly forbidden, which is a bad practice. Instead, it should be overloaded to raise an exception when appropriate, so that the API remains compatible. In this way, it should be possible to use indifferently `Engine` or `EngineMultirobot` in simulator, `addSystem` instead of `initialized`...

Once this limitation would be alleviated for Simulator, it should be suite straightfoward to adapt `Viewer` to support it, even though it is not clear how to deal with multiple robot. Looping over systems coule be done as an inner loop or an outer loop. Indeed, at replay, a multi-robots system is just a set of Independent single-robot systems.

duburcqa:
The only feature that it is missing at the time being is multi-robot replay.

PR #754, opened 2024-04-08
[python/simulator] First-class multi-robot support.

20chupin:
Related to https://github.com/duburcqa/jiminy/issues/246

### DIFF
--- a/core/src/engine/engine.cc
+++ b/core/src/engine/engine.cc
@@ -919,7 +919,7 @@ namespace jiminy
         {
             return ::jiminy::saveToBinary(value);
         }
-    };
+    }
 
     void Engine::start(const std::map<std::string, Eigen::VectorXd> & qInit,
                        const std::map<std::string, Eigen::VectorXd> & vInit,
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -119,7 +119,7 @@ def build_robot_from_log(
         mesh_package_dirs: Sequence[str] = (),
         *, robot_name: str = ""
         ) -> jiminy.Robot:
-    """Build a robot from log. Create and initialize a robot from a single or
+    """Create and initialize a robot from a single or
     multi robot simulation. If the robot to be built is from a multi-robot
     simulation, its name needs to be provided.
     
@@ -164,11 +164,11 @@ def build_robot_from_log(
     try:
         pinocchio_model = log_constants[
             ".".join((robot_namespace, "pinocchio_model"))]
-    except KeyError as exc:
+    except KeyError as e:
         if robot_name == "":
             raise ValueError(
                 "No robot with empty name has been found. Specify a name or "
-                "call `build_robots_from_log` to build the robot.") from exc
+                "call `build_robots_from_log`.") from e
 
     try:
         # Extract geometry models
@@ -259,7 +259,6 @@ def build_robots_from_log(
     log_constants = log_data["constants"]
 
     robots_names = []
-
     for key in log_constants.keys():
         if key == "HighLevelController.has_freeflyer":
             robots_names.append("")
@@ -269,10 +268,9 @@ def build_robots_from_log(
                 robots_names.append(m[0])
 
     robots = []
-
     for robot_name in robots_names:
-        robot = build_robot_from_log(log_data, mesh_path_dir,
-                                     mesh_package_dirs, robot_name=robot_name)
+        robot = build_robot_from_log(
+            log_data, mesh_path_dir, mesh_package_dirs, robot_name=robot_name)
 
         robots.append(robot)
 
@@ -373,9 +371,7 @@ def extract_trajectories_from_log(log_data: Dict[str, Any],
                    Optional: None by default. If None, then it will be
                    reconstructed from 'log_data' using `build_robots_from_log`.
 
-    :returns: Trajectory dictionary. The actual trajectory corresponds to the
-              field "evolution_robot" and it is a list of State object. The
-              other fields are additional information.
+    :returns: Dictonary mapping each robot name to its corresponding trajectory.
     """
     trajectories = {}
 
@@ -388,7 +384,6 @@ def extract_trajectories_from_log(log_data: Dict[str, Any],
             log_data, robot, robot_name=robot.name)
 
         trajectories[robot.name] = trajectory
-
     return trajectories
 
 def update_sensor_measurements_from_log(
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -146,8 +146,7 @@ def build_robot_from_log(
                               directories to the ones provided by log file. It
                               may be necessary to specify it to read log
                               generated on a different environment.
-    :param robot_name: Name of the robot to be constructed in the log. If it
-                       is not known, call `build_robots_from_log`.
+    :param robot_name: Name of the robot to build from log.
 
     :returns: Reconstructed robot, and parsed log data as returned by
               `jiminy_py.log.read_log` method.
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -122,7 +122,7 @@ def build_robot_from_log(
     """Create and initialize a robot from a single or
     multi robot simulation. If the robot to be built is from a multi-robot
     simulation, its name needs to be provided.
-    
+
     .. note::
         model options and `robot.pinocchio_model` will be the same as during
         the simulation until the next call to `reset` method unless the options
@@ -146,7 +146,7 @@ def build_robot_from_log(
                               directories to the ones provided by log file. It
                               may be necessary to specify it to read log
                               generated on a different environment.
-    :param robot_name: Name of the robot to be constructed in the log. If it 
+    :param robot_name: Name of the robot to be constructed in the log. If it
                        is not known, call `build_robots_from_log`.
 
     :returns: Reconstructed robot, and parsed log data as returned by
@@ -233,6 +233,7 @@ def build_robot_from_log(
 
     return robot
 
+
 def build_robots_from_log(
         log_data: Dict[str, Any],
         mesh_path_dir: Optional[str] = None,
@@ -241,7 +242,7 @@ def build_robots_from_log(
     """Build all the robots in the log of the simulation.
 
     .. note::
-        This function calls `build_robot_from_log` to build each robot. 
+        This function calls `build_robot_from_log` to build each robot.
         Refer to `build_robot_from_log` for more information.
 
     :param log_data: Logged data (constants and variables) as a dictionary.
@@ -294,7 +295,7 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
     :param robot: Jiminy robot associated with the logged trajectory.
                   Optional: None by default. If None, then it will be
                   reconstructed from 'log_data' using `build_robot_from_log`.
-    :param robot_name: Name of the robot to be constructed in the log. If it 
+    :param robot_name: Name of the robot to be constructed in the log. If it
                        is not known, call `build_robot_from_log`.
 
     :returns: Trajectory dictionary. The actual trajectory corresponds to the
@@ -352,11 +353,13 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
 
     return {"evolution_robot": evolution_robot,
             "robot": robot,
-            "use_theoretical_model": False}                  
+            "use_theoretical_model": False}
 
-def extract_trajectories_from_log(log_data: Dict[str, Any],
-                                robots: Optional[Sequence[jiminy.Model]] = None
-                                ) -> Dict[str, TrajectoryDataType]:
+
+def extract_trajectories_from_log(
+        log_data: Dict[str, Any],
+        robots: Optional[Sequence[jiminy.Model]] = None
+        ) -> Dict[str, TrajectoryDataType]:
     """Extract the minimal required information from raw log data in order to
     replay the simulation in a viewer.
 
@@ -371,7 +374,8 @@ def extract_trajectories_from_log(log_data: Dict[str, Any],
                    Optional: None by default. If None, then it will be
                    reconstructed from 'log_data' using `build_robots_from_log`.
 
-    :returns: Dictonary mapping each robot name to its corresponding trajectory.
+    :returns: Dictonary mapping each robot name to its corresponding
+              trajectory.
     """
     trajectories = {}
 
@@ -385,6 +389,7 @@ def extract_trajectories_from_log(log_data: Dict[str, Any],
 
         trajectories[robot.name] = trajectory
     return trajectories
+
 
 def update_sensor_measurements_from_log(
         log_data: Dict[str, Any],
--- a/python/jiminy_py/src/jiminy_py/log.py
+++ b/python/jiminy_py/src/jiminy_py/log.py
@@ -117,8 +117,11 @@ def build_robot_from_log(
         log_data: Dict[str, Any],
         mesh_path_dir: Optional[str] = None,
         mesh_package_dirs: Sequence[str] = (),
+        *, robot_name: str = ""
         ) -> jiminy.Robot:
-    """Build robot from log.
+    """Build robot from log. This function can only be used for logs of
+    single-robot simulations. For multi-robot simulations, you have to use
+    `build_robots_from_log`.
 
     .. note::
         model options and `robot.pinocchio_model` will be the same as during
@@ -147,27 +150,107 @@ def build_robot_from_log(
     :returns: Reconstructed robot, and parsed log data as returned by
               `jiminy_py.log.read_log` method.
     """
-    return build_robots_from_log(log_data, mesh_path_dir, mesh_package_dirs)[0]
+    # Prepare robot namespace
+    if robot_name == "":
+        ROBOT_NAMESPACE = ENGINE_NAMESPACE
+    else:
+        ROBOT_NAMESPACE = ".".join((ENGINE_NAMESPACE, robot_name))
+
+    # Instantiate empty robot
+    robot = jiminy.Robot(robot_name)
+
+    # Extract log constants
+    log_constants = log_data["constants"]
+
+    # Extract common info
+    pinocchio_model = log_constants[
+        ".".join((ROBOT_NAMESPACE, "pinocchio_model"))]
+
+    try:
+        # Extract geometry models
+        collision_model = log_constants[
+            ".".join((ROBOT_NAMESPACE, "collision_model"))]
+        visual_model = log_constants[
+            ".".join((ROBOT_NAMESPACE, "visual_model"))]
+
+        # Initialize the model
+        robot.initialize(pinocchio_model, collision_model, visual_model)
+    except KeyError as e:
+        # Extract initialization arguments
+        urdf_data = log_constants[
+            ".".join((ROBOT_NAMESPACE, "urdf_file"))]
+        has_freeflyer = int(log_constants[
+            ".".join((ROBOT_NAMESPACE, "has_freeflyer"))])
+        mesh_package_dirs = [*mesh_package_dirs, *log_constants.get(
+            ".".join((ROBOT_NAMESPACE, "mesh_package_dirs")), ())]
+
+        # Make sure urdf data is available
+        if len(urdf_data) <= 1:
+            raise RuntimeError(
+                "Impossible to build robot. The log is not persistent and the "
+                "robot was not associated with a valid URDF file.") from e
+
+        # Write urdf data in temporary file
+        urdf_path = os.path.join(
+            tempfile.gettempdir(),
+            f"{next(tempfile._get_candidate_names())}.urdf")
+        with open(urdf_path, "xb") as f:
+            f.write(urdf_data.encode())
+
+        # Fix the mesh paths in the URDF model if requested
+        if mesh_path_dir is not None:
+            fixed_urdf_path = _fix_urdf_mesh_path(urdf_path, mesh_path_dir)
+            os.remove(urdf_path)
+            urdf_path = fixed_urdf_path
+
+        # Initialize model
+        robot.initialize(urdf_path, has_freeflyer, mesh_package_dirs)
+
+        # Delete temporary file
+        os.remove(urdf_path)
+
+        # Load the options
+        all_options = log_constants[".".join((ENGINE_NAMESPACE, "options"))]
+        # Load the options
+        if robot_name == "":
+            robot.set_options(all_options["robot"])
+        else:
+            robot.set_options(all_options[".".join((robot_name, "robot"))])
+
+        # Update model in-place.
+        # Note that `__setstate__` re-allocates memory instead of just calling
+        # the copy assignment operator. Although this is undesirable, there is
+        # no better way on Python side. Anyway, this is not an issue in this
+        # particular case since the robot has just been created, so nobody got
+        # references to pre-allocated data at this point.
+        robot.pinocchio_model.__setstate__(pinocchio_model.__getstate__())
+
+        # Allocate corresponding pinocchio data manually
+        pinocchio_data = pinocchio_model.createData()
+        robot.pinocchio_data.__setstate__(pinocchio_data.__getstate__())
+
+    return robot
 
 def build_robots_from_log(
         log_data: Dict[str, Any],
         mesh_path_dir: Optional[str] = None,
         mesh_package_dirs: Sequence[str] = (),
         ) -> Sequence[jiminy.Robot]:
-    """Build robots from log.
+    """Build robots from log. For mulit-robot simulations, build all the robots
+    in the log of the simulation.
 
     .. note::
         model options and `robot.pinocchio_model` will be the same as during
         the simulation until the next call to `reset` method unless the options
-        of the robot that has been restored are overwritten manually.
+        of the robots that has been restored are overwritten manually.
 
     .. note::
-        It returns a valid and fully initialized robot, that can be used to
-        perform new simulation if added to a Jiminy Engine, but the original
-        controller is lost.
+        It returns a sequence of valid and fully initialized robots, that can
+        be used to perform new simulation if added to a Jiminy Engine, but the
+        original controller is lost.
 
     .. warning::
-        It does ot require the original URDF file to exist, but the original
+        It does ot require the original URDF files to exist, but the original
         mesh paths (if any) must be valid since they are not bundle in the log
         archive for now.
 
@@ -180,13 +263,10 @@ def build_robots_from_log(
                               may be necessary to specify it to read log
                               generated on a different environment.
 
-    :returns: Reconstructed robot, and parsed log data as returned by
-              `jiminy_py.log.read_log` method.
+    :returns: Sequence of reconstructed robots.
     """
     # Extract log constants
     log_constants = log_data["constants"]
-
-    all_options = log_constants[".".join((ENGINE_NAMESPACE, "options"))]
 
     robots_names = []
     
@@ -201,77 +281,8 @@ def build_robots_from_log(
     robots = []
 
     for robot_name in robots_names:
-        if robot_name == "":
-            ROBOT_NAMESPACE = ENGINE_NAMESPACE
-        else:
-            ROBOT_NAMESPACE = ".".join((ENGINE_NAMESPACE, robot_name))
-        # Instantiate empty robot
-        robot = jiminy.Robot(robot_name)
-
-        # Extract common info
-        pinocchio_model = log_constants[
-            ".".join((ROBOT_NAMESPACE, "pinocchio_model"))]
-
-        try:
-            # Extract geometry models
-            collision_model = log_constants[
-                ".".join((ROBOT_NAMESPACE, "collision_model"))]
-            visual_model = log_constants[
-                ".".join((ROBOT_NAMESPACE, "visual_model"))]
-
-            # Initialize the model
-            robot.initialize(pinocchio_model, collision_model, visual_model)
-        except KeyError as e:
-            # Extract initialization arguments
-            urdf_data = log_constants[
-                ".".join((ROBOT_NAMESPACE, "urdf_file"))]
-            has_freeflyer = int(log_constants[
-                ".".join((ROBOT_NAMESPACE, "has_freeflyer"))])
-            mesh_package_dirs = [*mesh_package_dirs, *log_constants.get(
-                ".".join((ROBOT_NAMESPACE, "mesh_package_dirs")), ())]
-
-            # Make sure urdf data is available
-            if len(urdf_data) <= 1:
-                raise RuntimeError(
-                    "Impossible to build robot. The log is not persistent and the "
-                    "robot was not associated with a valid URDF file.") from e
-
-            # Write urdf data in temporary file
-            urdf_path = os.path.join(
-                tempfile.gettempdir(),
-                f"{next(tempfile._get_candidate_names())}.urdf")
-            with open(urdf_path, "xb") as f:
-                f.write(urdf_data.encode())
-
-            # Fix the mesh paths in the URDF model if requested
-            if mesh_path_dir is not None:
-                fixed_urdf_path = _fix_urdf_mesh_path(urdf_path, mesh_path_dir)
-                os.remove(urdf_path)
-                urdf_path = fixed_urdf_path
-
-            # Initialize model
-            robot.initialize(urdf_path, has_freeflyer, mesh_package_dirs)
-
-            # Delete temporary file
-            os.remove(urdf_path)
-
-            # Load the options
-            if robot_name == "":
-                robot.set_options(all_options["robot"])
-            else:
-                robot.set_options(all_options[".".join((robot_name, "robot"))])
-
-            # Update model in-place.
-            # Note that `__setstate__` re-allocates memory instead of just calling
-            # the copy assignment operator. Although this is undesirable, there is
-            # no better way on Python side. Anyway, this is not an issue in this
-            # particular case since the robot has just been created, so nobody got
-            # references to pre-allocated data at this point.
-            robot.pinocchio_model.__setstate__(pinocchio_model.__getstate__())
-
-            # Allocate corresponding pinocchio data manually
-            pinocchio_data = pinocchio_model.createData()
-            robot.pinocchio_data.__setstate__(pinocchio_data.__getstate__())
+        robot = build_robot_from_log(log_data, mesh_path_dir,
+                                     mesh_package_dirs, robot_name=robot_name)
 
         robots.append(robot)
 
@@ -279,7 +290,8 @@ def build_robots_from_log(
 
 
 def extract_trajectory_from_log(log_data: Dict[str, Any],
-                                robot: Optional[jiminy.Model] = None
+                                robot: Optional[jiminy.Model] = None,
+                                *, robot_name: str = ""
                                 ) -> TrajectoryDataType:
     """Extract the minimal required information from raw log data in order to
     replay the simulation in a viewer.
@@ -299,8 +311,51 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
               field "evolution_robot" and it is a list of State object. The
               other fields are additional information.
     """
-    return extract_trajectories_from_log(log_data, [robot])[robot.name]
-                                
+    # Prepare robot namespace
+    if robot.name == "":
+        ROBOT_NAMESPACE = ENGINE_NAMESPACE
+    else:
+        ROBOT_NAMESPACE = ".".join((ENGINE_NAMESPACE, robot.name))
+    # Handling of default argument(s)
+    if robot is None:
+        robot = build_robot_from_log(log_data)
+
+    # Define some proxies for convenience
+    log_vars = log_data["variables"]
+
+    # Extract the joint positions, velocities and external forces over time
+    positions = np.stack([
+        log_vars.get(".".join((ROBOT_NAMESPACE, field)), [])
+        for field in robot.log_position_fieldnames], axis=-1)
+    velocities = np.stack([
+        log_vars.get(".".join((ROBOT_NAMESPACE, field)), [])
+        for field in robot.log_velocity_fieldnames], axis=-1)
+    forces = np.stack([
+        log_vars.get(".".join((ROBOT_NAMESPACE, field)), [])
+        for field in robot.log_f_external_fieldnames], axis=-1)
+
+    # Determine which optional data are available
+    has_positions = len(positions) > 0
+    has_velocities = len(velocities) > 0
+    has_forces = len(forces) > 0
+
+    # Create state sequence
+    evolution_robot = []
+    q, v, f_ext = None, None, None
+    for i, t in enumerate(log_vars["Global.Time"]):
+        if has_positions:
+            q = positions[i]
+        if has_velocities:
+            v = velocities[i]
+        if has_forces:
+            f_ext = [forces[i, (6 * (j - 1)):(6 * j)]
+                     for j in range(1, robot.pinocchio_model.njoints)]
+        evolution_robot.append(State(
+            t=t, q=q, v=v, f_ext=f_ext))  # type: ignore[arg-type]
+
+    return {"evolution_robot": evolution_robot,
+            "robot": robot,
+            "use_theoretical_model": False}                  
 
 def extract_trajectories_from_log(log_data: Dict[str, Any],
                                 robots: Optional[Sequence[jiminy.Model]] = None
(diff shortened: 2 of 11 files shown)


python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -146,76 +147,135 @@ def build_robot_from_log(
     :returns: Reconstructed robot, and parsed log data as returned by
               `jiminy_py.log.read_log` method.
     """
-    # Instantiate empty robot
-    robot = jiminy.Robot()
+    return build_robots_from_log(log_data, mesh_path_dir, mesh_package_dirs)[0]
```
duburcqa:
```python
robot, = build_robots_from_log(log_data, mesh_path_dir, mesh_package_dirs)
return return
```
This way, it would raise an exception of there is more than one robot. Update the doc to explain that it only works for single-robot simulations.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -146,76 +147,135 @@ def build_robot_from_log(
...
+        mesh_package_dirs: Sequence[str] = (),
+        ) -> Sequence[jiminy.Robot]:
+    """Build robots from log.
+
+    .. note::
+        model options and `robot.pinocchio_model` will be the same as during
+        the simulation until the next call to `reset` method unless the options
+        of the robot that has been restored are overwritten manually.
+
+    .. note::
```
duburcqa: Update the doc.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -146,76 +147,135 @@ def build_robot_from_log(
...
+    :param log_data: Logged data (constants and variables) as a dictionary.
+    :param mesh_path_dir: Overwrite the common root of all absolute mesh paths.
+                          It which may be necessary to read log generated on a
+                          different environment.
+    :param mesh_package_dirs: Prepend custom mesh package search path
+                              directories to the ones provided by log file. It
+                              may be necessary to specify it to read log
+                              generated on a different environment.
+
+    :returns: Reconstructed robot, and parsed log data as returned by
```
duburcqa: same.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -146,76 +147,135 @@ def build_robot_from_log(
...
+        model options and `robot.pinocchio_model` will be the same as during
+        the simulation until the next call to `reset` method unless the options
+        of the robot that has been restored are overwritten manually.
+
+    .. note::
+        It returns a valid and fully initialized robot, that can be used to
+        perform new simulation if added to a Jiminy Engine, but the original
+        controller is lost.
+
+    .. warning::
```
duburcqa: same.

python/jiminy_py/src/jiminy_py/log.py
duburcqa: Plus remove duplicated doc from `build_robot_from_log`, stating that it is a convenience wrapper over `build_robots_from_log` in case of single robot simulation, otherwise it is confusing.

duburcqa: After think twice, I would rather suggest to add and optional keyword-only argument `robot_name` to `build_robot_from_log`, and refactor `build_robots_from_log ` to simply call `build_robot_from_log` multiple times with different robot names:
```
def build_robot_from_log(
        log_data: Dict[str, Any],
        mesh_path_dir: Optional[str] = None,
        mesh_package_dirs: Sequence[str] = (),
        *, robot_name: str = ""
        ) -> jiminy.Robot:
```

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -239,47 +299,79 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
               field "evolution_robot" and it is a list of State object. The
               other fields are additional information.
     """
+    return extract_trajectories_from_log(log_data, [robot])[robot.name]
+                                
+
+def extract_trajectories_from_log(log_data: Dict[str, Any],
```
duburcqa: same. keep `extract_trajectory_from_log` as is, and adapt `extract_trajectories_from_log` to call `extract_trajectory_from_log ` multiple times and aggregate the outputs.

python/jiminy_py/src/jiminy_py/robot.py
```diff
@@ -847,6 +849,75 @@ def load_hardware_description_file(
...
 
+def build_robot(name: str,
+                urdf_path: str,
+                hardware_path: Optional[str] = None,
+                mesh_path_dir: Optional[str] = None,
+                has_freeflyer: bool = True,
+                avoid_instable_collisions: bool = True,
+                debug: bool = False) -> None:
+    """
+    Build a BaseJiminyRobot from urdf.
```
duburcqa: Explain why it is different from just calling `robot.initialize`. Currently this description is not providing any value to the user.

python/jiminy_py/src/jiminy_py/robot.py
```diff
@@ -847,6 +849,75 @@ def load_hardware_description_file(
     return extra_info
 
 
+def build_robot(name: str,
```
duburcqa: After thinking twice, move this method in `simulator.py` and call it `_build_robot_from_urdf`. What it does it non-standard and contradicts what `BaseJiminyRobot` is doing. For now it is not clear whether or not it should be made the default, so better limit its scope.

python/jiminy_py/src/jiminy_py/viewer/__init__.py
```diff
@@ -4,7 +4,8 @@
                      Viewer,
                      sleep,
                      is_display_available,
-                     get_default_backend)
+                     get_default_backend,
+                     COLORS)
```
duburcqa: Import globals first, better everything else.

python/jiminy_py/src/jiminy_py/viewer/__init__.py
```diff
@@ -25,5 +26,6 @@
     'play_trajectories',
     'play_logs_data',
     'play_logs_files',
-    'async_play_and_record_logs_files'
+    'async_play_and_record_logs_files',
+    'COLORS'
```
duburcqa: same.

python/jiminy_py/unit_py/test_simulator.py
```diff
@@ -74,6 +79,65 @@ def test_single_robot_simulation(self):
             robot, log_data, record_video_path=video_path, verbose=False)
         self.assertTrue(os.path.isfile(video_path))
 
+    def test_double_robot_simulation(self):
+        '''
```
duburcqa:
```python
"""Test simulation with two robots.
"""
```

python/jiminy_py/unit_py/test_simulator.py
```diff
@@ -74,6 +79,65 @@ def test_single_robot_simulation(self):
...
+
+        # Test: viewer
+        with self.assertRaises(NotImplementedError):
+            simulator.replay(verbose=False)
+
+        # Test: log reading
+        log_data = read_log(log_path)
+        trajectories = extract_trajectories_from_log(log_data)
+
+        np.testing.assert_array_almost_equal(simulator.robot_states[0].q, trajectories[simulator.robots[0].name]['evolution_robot'][-1].q, decimal=10)
```
duburcqa: Store `simulator.robots[i].name` and `simulator.robot_states[i]` in some variables to avoid infinite lines:
```python
robot_name_1, robot_name_2 = (robot.name for robot in simulator.robots)
robot_states_1, robot_states_2 = simulator.robot_states
```

python/jiminy_pywrap/src/engine.cc
```diff
@@ -277,16 +277,24 @@ namespace jiminy::python
             {
                 aInit.emplace(convertFromPython<std::map<std::string, Eigen::VectorXd>>(aInitPy));
             }
-            std::optional<AbortSimulationFunction> callback;
+            AbortSimulationFunction callback;
```
duburcqa, line 280: You should not need this. Irrelevant to this MR I think.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -73,13 +73,6 @@ def __init__(self,  # pylint: disable=unused-argument
         # Backup the user arguments
         self.viewer_kwargs = deepcopy(viewer_kwargs or {})
 
-        # Handling of default argument(s)
-        if "robot_name" not in self.viewer_kwargs:
```
duburcqa, line 77: What did you do about this ? You should make sure that the user does not specify `viewer_kwargs["robot_name"]` in case of multi-robot simulation (nor anything that is inherently single robot, I don't know if there are others.

Also, `robot.pinocchio_model.name` is not the same as `robot.name`. It is not clear to me which one to prefer. I would say most probably `robot.name`. Adding the robot suffix was necessary to make sure the name is unique. I'm not sure to what extend removing this will cause issues.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -206,6 +172,59 @@ def remove_file_at_exit(file_path: str) -> None:
...
 
+    def add_robot(self,
+                  name: str,
+                  urdf_path: str,
+                  hardware_path: Optional[str] = None,
+                  mesh_path_dir: Optional[str] = None,
+                  has_freeflyer: bool = True,
+                  avoid_instable_collisions: bool = True,
+                  debug: bool = False) -> None:
+        r"""
```
duburcqa: Add doc. No `r` needed here.

python/jiminy_py/src/jiminy_py/robot.py
```diff
@@ -847,6 +849,75 @@ def load_hardware_description_file(
...
+                urdf_path: str,
+                hardware_path: Optional[str] = None,
+                mesh_path_dir: Optional[str] = None,
+                has_freeflyer: bool = True,
+                avoid_instable_collisions: bool = True,
+                debug: bool = False) -> None:
```
duburcqa: `None` ?

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -206,6 +172,59 @@ def remove_file_at_exit(file_path: str) -> None:
...
+                                          box, and replacing primitive box by
+                                          its vertices.
+        :param debug: Whether the debug mode must be activated. Doing it
+                      enables temporary files automatic deletion.
+        """
+        robot = build_robot(name, urdf_path, hardware_path, mesh_path_dir,
+                            has_freeflyer, avoid_instable_collisions, debug)
+
+        # Get engine options
+        engine_options = self.engine.get_options()
```
duburcqa, line 327: just do a for-loop:
```python
for hardware_path, option_nested_path in (
        ('groundStiffness', ('contacts', 'stiffness')), 
        ('groundDamping',), ('contacts', 'damping')),
        ('controllerUpdatePeriod',), ('stepper', 'controllerUpdatePeriod'))):
    if hardware_path in robot.extra_info:
        option = engine_options
        for option_path in option_nested_path:
            option = option[option_path]
        if robot.extra_info[hardware_path] != option:
           warnings.warn(...)
```

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -229,6 +248,15 @@ def __dir__(self) -> List[str]:
...
         """
         return [*super().__dir__(), *dir(self.engine)]
+    
+    @property
+    def viewer(self) -> Viewer:
+        """
+        """
+        if len(self._viewers) == 0:
+            return None
+        else:
```
duburcqa: No else needed here

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -229,6 +248,15 @@ def __dir__(self) -> List[str]:
...
+            return self._viewers[0]
```
duburcqa, line 372: Check that the simulation is single robot. Raise an exception otherwise.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -229,6 +248,15 @@ def __dir__(self) -> List[str]:
         to get consistent autocompletion wrt `getattr`.
         """
         return [*super().__dir__(), *dir(self.engine)]
+    
+    @property
+    def viewer(self) -> Viewer:
+        """
```
duburcqa, line 370: Add doc. State that it only works for single robot.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -73,13 +73,6 @@ def __init__(self,  # pylint: disable=unused-argument
         # Backup the user arguments
```
duburcqa, line 167: Update simulator doc to explain that it is now multi-robot, but it has been design to remain as easy of use as possible for single-robot simulation, since it is by far the most common use-case.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -229,6 +248,15 @@ def __dir__(self) -> List[str]:
...
+    @property
+    def viewer(self) -> Viewer:
+        """
+        """
+        if len(self._viewers) == 0:
+            return None
+        else:
+            return self._viewers[0]
 
     @property
```
duburcqa, line 382: Add a warning to all `property` methods stating that they only work in case of single robot simulation.
```python
robot, = self.engine.robots
return robot
```
This way it would crash for multi-robot simulations.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -255,7 +283,7 @@ def robot_state(self) -> jiminy.RobotState:
...
     def is_viewer_available(self) -> bool:
-        """Returns whether a viewer instance associated with the robot
+        """Returns whether a viewer instance associated with the first robot
```
duburcqa: `associated with the ongoing simulation.` should be fine, and closer to the question the user is trying to answer.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -294,6 +322,12 @@ def register_profile_force(self,
                               the whole simulation. There is no issue for C++
                               bindings such as `jiminy.RandomPerlinProcess`.
         """
+        if len(self.engine.robots) > 1:
+            raise NotImplementedError("To register a force in multirobot "
```
duburcqa: To avoid so many lines, rather do:
```python
raise NotImplementedError(
    "......."
    ".......")
```

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -316,6 +350,12 @@ def register_impulse_force(self,
                            is aligned with world frame. It is represented as a
                            `np.ndarray` (Fx, Fy, Fz, Mx, My, Mz).
         """
+        if len(self.engine.robots) > 1:
+            raise NotImplementedError("To register a force in multirobot "
```
duburcqa: same

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -357,36 +397,50 @@ def reset(self, remove_all_forces: bool = False) -> None:
...
+        :param v_init: Initial velocity (by robot if it is a dictionnary).
+        :param a_init: Initial acceleration (by robot if it is a dictionnary).
+                       It is only used by acceleration dependent sensors and
+                       controllers, such as IMU and force sensors.
+        :param is_state_theoretical: In single robot simulations, whether the initial state is associated
                                      with the actual or theoretical model of
                                      the robot.
         """
         # Call base implementation
-        self.engine.start(q_init, v_init, a_init, is_state_theoretical)
```
duburcqa, line 376: This method works for single-robot simulations, no need for anything fancy.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -357,36 +397,50 @@ def reset(self, remove_all_forces: bool = False) -> None:
...
+                v_init = v_init[list(v_init.keys())[0]]
+                if a_init:
+                    a_init = a_init[list(a_init.keys())[0]]
+            if is_state_theoretical is None:
+                is_state_theoretical = False
+            self.engine.start(q_init, v_init, a_init, is_state_theoretical)
+        else:
+            if is_state_theoretical is not None:
+                raise Exception(
+                    "`is_state_theoretical` can only be precised in single "
```
duburcqa: `is only supported for` to stick to the usual terminology elsewhere.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -357,36 +397,50 @@ def reset(self, remove_all_forces: bool = False) -> None:
...
+                    "robot simulations.")
+            self.engine.start(q_init, v_init, a_init)
 
         # Share the external force buffer of the viewer with the engine.
         # Note that the force vector must be converted to pain list to avoid
         # copy with external sub-vector.
-        if self.viewer is not None:
-            self.viewer.f_external = [*self.robot_state.f_external][1:]
+        for i, viewer in enumerate(self._viewers):
```
duburcqa: zip would be better, as it also may sure that it will have the appropriate length, ie the min between the number of elements of `robot_states` and `self._viewers`:
```python
for robot_state, viewer in zip(self.robot_states, self._viewers):
    ....
```

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -519,59 +590,67 @@ def render(self,
...
-            # Initialize camera pose
-            if self.viewer.is_backend_parent and camera_pose is None:
-                camera_pose = viewer_kwargs.get("camera_pose", (
-                    (9.0, 0.0, 2e-5), (np.pi/2, 0.0, np.pi/2), None))
+        for i, viewer in enumerate(self._viewers):
+            if not viewer.is_open():
+                viewer.close()
+                self._viewers.pop(i)
+
+        for i, robot in enumerate(self.robots):
```
duburcqa: So many indents x) Maybe at this point  you should add yet another private helper method on top of this module `_create_viewer`.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -519,59 +590,67 @@ def render(self,
...
+                        self._viewers[-1].display_external_forces(visibility)
+
+                # Initialize camera pose
+                if self._viewers[-1].is_backend_parent and camera_pose is None:
+                    camera_pose = viewer_kwargs.get("camera_pose", (
+                        (9.0, 0.0, 2e-5), (np.pi/2, 0.0, np.pi/2), None))
 
         # Enable the ground profile is requested and available
-        assert self.viewer is not None and self.viewer.backend is not None
+        assert self._viewers is not [] and [viewer.backend is not None for viewer in self._viewers]
```
duburcqa: `assert self._viewers and Viewer.backend is not None` should be fine

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -669,8 +752,9 @@ def close(self) -> None:
         """Close the connection with the renderer.
         """
         if hasattr(self, "viewer") and self.viewer is not None:
-            self.viewer.close()
-            self.viewer = None
+            for viewer in self._viewers:
+                viewer.close()
+            self._viewers = []
```
duburcqa: `self._viewers.clear()`

python/jiminy_py/src/jiminy_py/simulator.py
20chupin: I think that if you want to precise `is_state_theoretical` you need to send lists. If I remove the cases I have :
```
Engine.start(Engine, dict, dict, NoneType, bool)
did not match C++ signature:
    start(jiminy::Engine {lvalue} self, Eigen::Matrix<double, -1, 1, 0, -1, 1> q_init, Eigen::Matrix<double, -1, 1, 0, -1, 1> v_init, boost::python::api::object a_init=None, bool is_state_theoretical=False)
    start(jiminy::Engine {lvalue} self, boost::python::dict q_init_dict, boost::python::dict v_init_dict, boost::python::api::object a_init_dict=None)
```

python/jiminy_pywrap/src/engine.cc
20chupin: I did this change because the callback was not sent to `simulate`. The progress bar did not progress. If I revert this change and I still want to send the callback, I have :
```
error: no matching function for call to ‘jiminy::Engine::simulate(double&, std::enable_if_t<true, std::map<std::__cxx11::basic_string<char>, Eigen::Matrix<double, -1, 1>, std::less<std::__cxx11::basic_string<char> >, std::allocator<std::pair<const std::__cxx11::basic_string<char>, Eigen::Matrix<double, -1, 1> > > > >, std::enable_if_t<true, std::map<std::__cxx11::basic_string<char>, Eigen::Matrix<double, -1, 1>, std::less<std::__cxx11::basic_string<char> >, std::allocator<std::pair<const std::__cxx11::basic_string<char>, Eigen::Matrix<double, -1, 1> > > > >, std::optional<std::map<std::__cxx11::basic_string<char>, Eigen::Matrix<double, -1, 1>, std::less<std::__cxx11::basic_string<char> >, std::allocator<std::pair<const std::__cxx11::basic_string<char>, Eigen::Matrix<double, -1, 1> > > > >&, std::optional<std::function<bool()> >&)’
  285 |             return self.simulate(
      |                    ~~~~~~~~~~~~~^
  286 |                 endTime,
      |                 ~~~~~~~~         
  287 |                 convertFromPython<std::map<std::string, Eigen::VectorXd>>(qInitPy),
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  288 |                 convertFromPython<std::map<std::string, Eigen::VectorXd>>(vInitPy),
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  289 |                 aInit,
      |                 ~~~~~~           
  290 |                 callback);
      |                 ~~~~~~~~~        
compilation terminated due to -Wfatal-errors.
```

core/src/engine/engine.cc
```diff
@@ -919,7 +919,7 @@ namespace jiminy
         {
             return ::jiminy::saveToBinary(value);
         }
-    };
```
matthieuvigne: Is this intended ? This seems unvalid...

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -239,47 +299,79 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
               field "evolution_robot" and it is a list of State object. The
               other fields are additional information.
     """
+    return extract_trajectories_from_log(log_data, [robot])[robot.name]
```
matthieuvigne: Does this really work if `robot`  is None ? I guess it does because `robot` is updated through `extract_trajectories_from_log` - then I would suggest writing it in two lines (like `build_robot_from_log`) for readability (possibly doing the same thing that @duburcqa suggested for this function).

matthieuvigne (approved):

core/src/engine/engine.cc
duburcqa: Should disappear after a rebase. It was a "bug" in jiminy master, after merging something a bit too fast.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -116,8 +117,11 @@ def build_robot_from_log(
         log_data: Dict[str, Any],
         mesh_path_dir: Optional[str] = None,
         mesh_package_dirs: Sequence[str] = (),
+        *, robot_name: str = ""
         ) -> jiminy.Robot:
-    """Build robot from log.
+    """Build robot from log. This function can only be used for logs of
```
duburcqa: From what I seeing. It can actually be used for multi-robot simulations. Just that it can only load one specific robot, providing its name.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -116,8 +117,11 @@ def build_robot_from_log(
         log_data: Dict[str, Any],
         mesh_path_dir: Optional[str] = None,
         mesh_package_dirs: Sequence[str] = (),
+        *, robot_name: str = ""
```
duburcqa, line 120: Add doc for this parameter.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -146,33 +150,39 @@ def build_robot_from_log(
     :returns: Reconstructed robot, and parsed log data as returned by
               `jiminy_py.log.read_log` method.
     """
+    # Prepare robot namespace
+    if robot_name == "":
+        ROBOT_NAMESPACE = ENGINE_NAMESPACE
+    else:
+        ROBOT_NAMESPACE = ".".join((ENGINE_NAMESPACE, robot_name))
```
duburcqa: `ROBOT_NAMESPACE = ".".join(filter(None, (ENGINE_NAMESPACE, robot_name)))` should work in both cases.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -201,7 +211,11 @@ def build_robot_from_log(
 
         # Load the options
         all_options = log_constants[".".join((ENGINE_NAMESPACE, "options"))]
-        robot.set_options(all_options["robot"])
+        # Load the options
+        if robot_name == "":
+            robot.set_options(all_options["robot"])
+        else:
+            robot.set_options(all_options[".".join((robot_name, "robot"))])
```
duburcqa: same. Use `filter`.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -217,9 +231,67 @@ def build_robot_from_log(
 
     return robot
 
+def build_robots_from_log(
+        log_data: Dict[str, Any],
+        mesh_path_dir: Optional[str] = None,
+        mesh_package_dirs: Sequence[str] = (),
+        ) -> Sequence[jiminy.Robot]:
+    """Build robots from log. For mulit-robot simulations, build all the robots
```
duburcqa: In any case, it just "Build all the robots in the log of the simulation."

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -217,9 +231,67 @@ def build_robot_from_log(
...
+    in the log of the simulation.
+
+    .. note::
```
duburcqa, line 243: Remove redundant notes and refer to `build_robot_from_log` for more information.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -217,9 +231,67 @@ def build_robot_from_log(
...
+
+        robots.append(robot)
+
+    return robots
+
 
 def extract_trajectory_from_log(log_data: Dict[str, Any],
-                                robot: Optional[jiminy.Model] = None
+                                robot: Optional[jiminy.Model] = None,
+                                *, robot_name: str = ""
```
duburcqa: No need for `robot_name` if the robot has been specified, but you need it if it was not...
So I would suggest to change the signature for this: `robot_name: Optional[str] = None`, and use `robot.name` by default if specified, or `""` if not and `robot_name` is not specified either, and finally `robot_name` if specified by not robot. If both are specified, just raise an exception.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -278,8 +355,41 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
...
-            "use_theoretical_model": False}
+            "use_theoretical_model": False}                  
+
+def extract_trajectories_from_log(log_data: Dict[str, Any],
+                                robots: Optional[Sequence[jiminy.Model]] = None
+                                ) -> Dict[str, TrajectoryDataType]:
+    """Extract the minimal required information from raw log data in order to
+    replay the simulation in a viewer.
+
+    .. note::
```
duburcqa, line 365: Do no duplicate doc. Just refer to the other method.

python/jiminy_py/src/jiminy_py/robot.py
```diff
@@ -866,11 +866,11 @@ class BaseJiminyRobot(jiminy.Robot):
         name than the URDF file will be detected automatically without
         requiring to manually specify its path.
     """
-    def __init__(self) -> None:
+    def __init__(self, name : Optional[str] = "") -> None:
```
duburcqa: Add documentation for this new parameter.

python/jiminy_py/unit_py/test_simulator.py
```diff
@@ -64,7 +65,10 @@ def test_single_robot_simulation(self):
 
         # Test: log reading
         log_data = read_log(log_path)
-        robot = build_robot_from_log(log_data)
+        trajectory = extract_trajectory_from_log(log_data, robot)
+
+        np.testing.assert_array_almost_equal(simulator.robot_state.q, trajectory['evolution_robot'][-1].q, decimal=10)
```
duburcqa: line too long.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -278,8 +355,41 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
...
+        It extracts the required data for replay, namely temporal evolution of:
+          - robot configuration: to display of the robot on the scene,
+          - robot velocity: to update velocity-dependent markers such as DCM,
+          - external forces: to update force-dependent markers.
+
+    :param log_data: Logged data (constants and variables) as a dictionary.
+    :param robot: Jiminy robot associated with the logged trajectory.
```
duburcqa: Wrong documentation.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -47,14 +49,91 @@
...
+        avoid_instable_collisions, load_visual_meshes=debug, verbose=debug)
+
+    return robot
+
 class Simulator:
-    """Single-robot simulation wrapper providing a unified user-API on top of
+    """Simulation wrapper providing a unified user-API on top of
     the low-level jiminy C++ core library and Python-native modules for 3D
-    rendering and log data visualization.
+    rendering and log data visualization. The simulation can now be multi-robot
```
duburcqa: Add more details about how it works under the hood (empty robot name by default plus its special meaning...etc). Explain that methods that refer to a single robot are systematically returning info associated with the first robot, no matter if the simulation is single or multi robot... Explain what is not supported by multi-robot... A lot of info are missing here

python/jiminy_py/src/jiminy_py/log.py
20chupin: The method extract_trajectories_from_log is now calling extract_trajectory_from_log, so it should have disappeared.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -142,37 +146,47 @@ def build_robot_from_log(
...
     # Extract log constants
     log_constants = log_data["constants"]
 
-    # Extract common info
-    pinocchio_model = log_constants[
-        ".".join((ENGINE_NAMESPACE, "pinocchio_model"))]
+    try:
+        pinocchio_model = log_constants[
+            ".".join((robot_namespace, "pinocchio_model"))]
+    except KeyError as exc:
```
duburcqa: `e` is more common that `exc`

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -142,37 +146,47 @@ def build_robot_from_log(
...
+        if robot_name == "":
+            raise ValueError(
```
duburcqa, line 168: Remove ` to build the robot.`.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -116,9 +117,12 @@ def build_robot_from_log(
         log_data: Dict[str, Any],
         mesh_path_dir: Optional[str] = None,
         mesh_package_dirs: Sequence[str] = (),
+        *, robot_name: str = ""
         ) -> jiminy.Robot:
-    """Build robot from log.
-
+    """Build a robot from log. Create and initialize a robot from a single or
```
duburcqa: Remove `Build a robot from log.`

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -217,9 +233,55 @@ def build_robot_from_log(
...
+                              may be necessary to specify it to read log
+                              generated on a different environment.
+
+    :returns: Sequence of reconstructed robots.
+    """
+    # Extract log constants
+    log_constants = log_data["constants"]
+
+    robots_names = []
+
```
duburcqa: No extra space here.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -217,9 +233,55 @@ def build_robot_from_log(
...
+    for key in log_constants.keys():
+        if key == "HighLevelController.has_freeflyer":
+            robots_names.append("")
+        else:
+            m = re.findall(r"HighLevelController.(\w+).has_freeflyer", key)
+            if len(m) == 1:
+                robots_names.append(m[0])
+
+    robots = []
+
```
duburcqa: same.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -217,9 +233,55 @@ def build_robot_from_log(
...
+    for robot_name in robots_names:
+        robot = build_robot_from_log(log_data, mesh_path_dir,
```
duburcqa: The following syntax should be preferred if it fits the character limit.
```
robot = build_robot_from_log(
    log_data, mesh_path_dir, mesh_package_dirs, robot_name=robot_name)
```

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -278,8 +354,42 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
...
+        robot's trajectory. Refer to `extract_trajectory_from_log` for more
+        information.
+
+    :param log_data: Logged data (constants and variables) as a dictionary.
+    :param robots: Sequend of Jiminy robots associated with the logged
+                   trajectory.
+                   Optional: None by default. If None, then it will be
+                   reconstructed from 'log_data' using `build_robots_from_log`.
+
+    :returns: Trajectory dictionary. The actual trajectory corresponds to the
```
duburcqa: Wrong doc.

duburcqa: Just say: `Dictonary` mapping  each robot name to its corresponding trajectory.

python/jiminy_py/src/jiminy_py/log.py
```diff
@@ -278,8 +354,42 @@ def extract_trajectory_from_log(log_data: Dict[str, Any],
...
+    trajectories = {}
+
+    # Handling of default argument(s)
+    if robots is None:
+        robots = build_robots_from_log(log_data)
+
+    for robot in robots:
+        trajectory = extract_trajectory_from_log(
+            log_data, robot, robot_name=robot.name)
+
```
duburcqa, line 388: no extra space.

python/jiminy_py/src/jiminy_py/robot.py
```diff
@@ -865,12 +865,14 @@ class BaseJiminyRobot(jiminy.Robot):
         Hardware description files within the same directory and having the
         name than the URDF file will be detected automatically without
         requiring to manually specify its path.
+
+    :param name: Name of the robot to be created, empty string by default.
```
duburcqa: Use the same convention as everywhere else:
```python
:param name: Name of the robot to be created.
             Optional: Empty string by default.
```

python/jiminy_py/unit_py/test_simulator.py
```diff
@@ -64,7 +65,14 @@ def test_single_robot_simulation(self):
 
         # Test: log reading
         log_data = read_log(log_path)
-        robot = build_robot_from_log(log_data)
+        trajectory = extract_trajectory_from_log(log_data, robot)
```
duburcqa: Do not specify `robot` for this test. It must be reconstructed.

python/jiminy_py/unit_py/test_simulator.py
```diff
@@ -74,6 +82,73 @@ def test_single_robot_simulation(self):
             robot, log_data, record_video_path=video_path, verbose=False)
         self.assertTrue(os.path.isfile(video_path))
 
+    def test_double_robot_simulation(self):
+        """Test simulation with two robots.
+        """
+        robot1_name = "robot1"
```
duburcqa: `robot1_name, robot2_name = "robot1", "robot2"`

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -47,14 +49,109 @@
...
 
+def _build_robot_from_urdf(name: str,
+                           urdf_path: str,
+                           hardware_path: Optional[str] = None,
+                           mesh_path_dir: Optional[str] = None,
+                           has_freeflyer: bool = True,
+                           avoid_instable_collisions: bool = True,
+                           debug: bool = False) -> jiminy.Robot:
+    r"""Create and initialize a new robot from scratch, based on configuration
+    files only. Create a temporary harware file if needed.
```
duburcqa: `Create a default harware file if none is provided. See `generate_default_hardware_description_file` for details.`

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -47,14 +49,109 @@
...
+        urdf_path, hardware_path, mesh_path_dir, (), has_freeflyer,
+        avoid_instable_collisions, load_visual_meshes=debug, verbose=debug)
+
+    return robot
+
 class Simulator:
-    """Single-robot simulation wrapper providing a unified user-API on top of
-    the low-level jiminy C++ core library and Python-native modules for 3D
-    rendering and log data visualization.
+    """Simulation wrapper providing a unified user-API on top of the low-level
```
duburcqa, line 123: This doc looks amazing, but it needs some formatting x) Don't be afraid to skip lines. See https://github.com/duburcqa/jiminy/blob/774be9a99db2f8051caedd67f17c5a9fb92273a4/python/gym_jiminy/common/gym_jiminy/common/blocks/deformation_estimator.py#L417 as an example.

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -206,6 +275,67 @@ def remove_file_at_exit(file_path: str) -> None:
...
 
+    def add_robot(self,
+                  name: str,
+                  urdf_path: str,
+                  hardware_path: Optional[str] = None,
+                  mesh_path_dir: Optional[str] = None,
+                  has_freeflyer: bool = True,
+                  avoid_instable_collisions: bool = True,
+                  debug: bool = False) -> None:
+        r"""Create and add a new robot to the simulator from scratch, based on
```
duburcqa: `Create a new robot from scratch based on configuration files only and add to the simulator.` is better no ?

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -230,37 +360,137 @@ def __dir__(self) -> List[str]:
         """
         return [*super().__dir__(), *dir(self.engine)]
 
+    def _create_viewer(self,
+                       robot: jiminy.Robot,
+                       robot_state: jiminy.RobotState,
+                       robot_color: Optional[
+                           Tuple[float, float, float, float]],
+                       **viewer_kwargs: Any) -> None:
+        """Helper method to create a viewer associated to robot i.
```
duburcqa: `associated with a given robot`

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -604,6 +856,10 @@ def replay(self,
         :param kwargs: Extra keyword arguments for delegation to
                        `replay.play_trajectories` method.
         """
+        if len(self.robots) > 1:
+            raise NotImplementedError("Simulator.replay() is only supported "
```
duburcqa: It does not fit like this ?
```python
            raise NotImplementedError(
                "Simulator.replay() is only supported for single-robot simulations.")
```

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -668,9 +924,10 @@ def replay(self,
     def close(self) -> None:
         """Close the connection with the renderer.
         """
-        if hasattr(self, "viewer") and self.viewer is not None:
-            self.viewer.close()
-            self.viewer = None
+        if hasattr(self, "_viewers") and len(self._viewers) > 0:
```
duburcqa: `if hasattr(self, "_viewers")` is enough.

python/jiminy_py/src/jiminy_py/simulator.py
20chupin: 86 characters ;(

python/jiminy_py/src/jiminy_py/simulator.py
20chupin: Much better !

python/jiminy_py/src/jiminy_py/simulator.py
20chupin: Hope you'll like the new format.

duburcqa:
Looks like it is ready to merge now  ;)

duburcqa:
Good job !

duburcqa:
Just make sure that all tests pass.

duburcqa (dismissed):

python/jiminy_py/src/jiminy_py/simulator.py
```diff
@@ -230,37 +373,137 @@ def __dir__(self) -> List[str]:
...
+                    robot.pinocchio_model.frames[f.frame_index].parent
+                    for f in impulse_forces)
+                visibility = self._viewers[-1]._display_f_external
+                assert isinstance(visibility, list)
+                for i in force_frames:
+                    visibility[i - 1] = True
+                self._viewers[-1].display_external_forces(visibility)
+
+    @property
+    def viewer(self) -> Optional[Viewer]:
```
duburcqa, line 367: Change the return type in favor of "Viewer" and raise an exception if none is available.

PR #774, opened 2024-04-23
[gym] Enable multi-GPU support with PPO when mirror_mat are defined

### LINKED ISSUE #773: [gym] Fix typing and docs RLlib utilities with ray >= 2.9
`algo.save()` doesn't return a checkpoint path anymore. Instead, we need to use `algo.save().checkpoint.path`. This directly impacts the `train` function in `gym_jiminy.rllib.utilities`

mwulfman:
Closes #773

### DIFF
--- a/python/gym_jiminy/rllib/gym_jiminy/rllib/ppo.py
+++ b/python/gym_jiminy/rllib/gym_jiminy/rllib/ppo.py
@@ -155,6 +155,10 @@ def _compute_mirrored_value(value: torch.Tensor,
     """Compute mirrored value from observation space based on provided
     mirroring transformation.
     """
+    # Make sure value and mirror_mat are on the same device.
+    # This is needed for multi-GPU training.
+    mirror_mat = mirror_mat.to(value.device)
+
     def _update_flattened_slice(data: torch.Tensor,
                                 shape: Tuple[int, ...],
                                 mirror_mat: torch.Tensor) -> torch.Tensor:


duburcqa:
You broke something else by trying to fix your bug. I suggest to keep the output of train as it is currently, but fix typing and doc, then adapt the other parts that assumes it returns the checkpoint path.

python/gym_jiminy/rllib/gym_jiminy/rllib/ppo.py
```diff
@@ -155,6 +155,11 @@ def _compute_mirrored_value(value: torch.Tensor,
     """Compute mirrored value from observation space based on provided
     mirroring transformation.
     """
+    # Make sure value and mirror_mat are on the same device.
+    # This is needed for multi-GPU training.
+    if value.device != mirror_mat.device:
+        mirror_mat = mirror_mat.to(value.device)
```
duburcqa: no need to check the device. it is a no-op if unecessary.

python/gym_jiminy/rllib/gym_jiminy/rllib/utilities.py
```diff
@@ -368,7 +369,7 @@ def train(algo: Algorithm,
           logdir: Optional[Logger] = None,
           checkpoint_period: int = 0,
           verbose: bool = True,
-          debug: bool = False) -> str:
+          debug: bool = False) -> _TrainingResult:
```
duburcqa: I already implemented a patch in my other PR. Maybe you could backport what I did here. I added some extra details.

python/gym_jiminy/rllib/gym_jiminy/rllib/ppo.py
mwulfman: Fixed it.

python/gym_jiminy/rllib/gym_jiminy/rllib/utilities.py
mwulfman: Ok, I reverted this change and renamed the PR.

PR #780, opened 2024-04-30
[gym] Add symmetry surrogate loss in rllib custom ppo

mwulfman:
This PR adds the possibility to add the surrogate loss on the mirrored observation-action training data. Naming can be changed if desired as we could name it also "symmetry_surrogate_loss" for instance.


duburcqa:
I'm in favour of changing the name of the option as you suggested. It would be nice if you could update to documentation to explain what it does and give the source of the article introducing this approach. Thank you !

mwulfman:
I've renamed the loss to `symmetry_surrogate_loss` and added some documentation to review. Should I add the reference article for CAPS as well ?

duburcqa:
If you are motivated, then it would be nice yes ! You could even add mine in addition that is giving more insights regarding the interesting properties that emerges and how to tune the parameters.

mwulfman:
I have updated the documentation to add the mentioned articles.

PR #797, opened 2024-05-17
[core] Add stairs terrain

mwulfman:
I am proposing to add a terrain with alternating up and down staircases illustrated on the screenshot below.

stairs_ground


duburcqa:
You can move it in `core/utilities/geometry.h`.

core/src/utilities/random.cc
```diff
@@ -735,4 +735,27 @@ namespace jiminy
...
+        const Eigen::Rotation2D<double> rot_mat(orientation);
+
+        return [stepWidth, stepHeight, stepsNumber, rot_mat](
+                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
+        {
+            // Compute position in stairs reference frame
+            Eigen::Vector2d posRel = (rot_mat.inverse() * pos).array();
+            auto modPos = std::fmod(std::abs(posRel[0]), stepWidth*stepsNumber*2);
+
+            auto stairNumero = static_cast<uint32_t>(modPos / stepWidth);
```
duburcqa: `stairIndex`

duburcqa: Add type explicitly if trivial. Mark `const` if not modified.

core/src/utilities/random.cc
```diff
@@ -735,4 +735,27 @@ namespace jiminy
...
+                             uint32_t stepsNumber,
+                             double orientation)
+    {
+        const Eigen::Rotation2D<double> rot_mat(orientation);
+
+        return [stepWidth, stepHeight, stepsNumber, rot_mat](
+                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
+        {
+            // Compute position in stairs reference frame
+            Eigen::Vector2d posRel = (rot_mat.inverse() * pos).array();
```
duburcqa: `.array()`is useless.

core/src/utilities/random.cc
```diff
@@ -735,4 +735,27 @@ namespace jiminy
...
+            auto modPos = std::fmod(std::abs(posRel[0]), stepWidth*stepsNumber*2);
+
+            auto stairNumero = static_cast<uint32_t>(modPos / stepWidth);
+            if (stairNumero > stepsNumber){
+                stairNumero = 2*stepsNumber - stairNumero;
```
duburcqa: spacing.

core/src/utilities/random.cc
```diff
@@ -735,4 +735,27 @@ namespace jiminy
...
+
+        return [stepWidth, stepHeight, stepsNumber, rot_mat](
+                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
+        {
+            // Compute position in stairs reference frame
+            Eigen::Vector2d posRel = (rot_mat.inverse() * pos).array();
+            auto modPos = std::fmod(std::abs(posRel[0]), stepWidth*stepsNumber*2);
+
+            auto stairNumero = static_cast<uint32_t>(modPos / stepWidth);
+            if (stairNumero > stepsNumber){
```
duburcqa: Apply clang format.

python/jiminy_pywrap/src/generators.cc
```diff
@@ -193,6 +193,7 @@ namespace jiminy::python
         bp::def("random_tile_ground", &tiles,
                                       (bp::arg("size"), "height_max", "interp_delta",
                                        "sparsity", "orientation", "seed"));
+        bp::def("stairs_ground", &stairs, (bp::arg("step_width"), "step_height", bp::arg("steps_number"), "orientation"));
```
duburcqa: `, "steps_number", `

core/include/jiminy/core/utilities/random.h
```diff
@@ -536,6 +536,11 @@ namespace jiminy
                                           uint32_t sparsity,
                                           double orientation,
                                           uint32_t seed);
+
+    HeightmapFunction JIMINY_DLLAPI stairs(double width,
```
duburcqa: Add documentation. With ASCII art of the ground profile.

duburcqa:
Please add screenshot.

core/src/utilities/random.cc
```diff
@@ -735,4 +735,27 @@ namespace jiminy
...
+                stairNumero = 2*stepsNumber - stairNumero;
+            }
+            height = stairNumero * stepHeight;
```
duburcqa: vertical edges are not supported. Add a small slope.

mwulfman:
> Please add screenshot.

Where should I add it? 
Screenshot from 2024-05-21 15-53-26

duburcqa:
Just edit the original PR description :)

mwulfman:
Ok perfect. I've taken into account the review, added documentation and ASCII art. Waiting for potential new remarks.

core/src/utilities/geometry.cc
```diff
@@ -749,4 +749,41 @@ namespace jiminy
...
+            height = stairIndex * stepHeight;
+            normal = Eigen::Vector3d::UnitZ();
+
+            // Avoid unsupported vertical edge
+            const double posRelOnStep = std::fmod(modPos, stepWidth) / stepWidth;
+            if (1 - posRelOnStep < interpDelta)
+            {
+                height += staircaseSlopeSign * (stepHeight / interpDelta) *
+                          (posRelOnStep - (1.0 - interpDelta));
+                normal << -staircaseSlopeSign * stepHeight / interpDelta, 0.0, 1.0;
```
duburcqa: Normalise by yourself at this point. It would be more efficient:
```cpp
const double slope = staircaseSlopeSign * stepHeight / interpDelta;
const double normInv = 1.0 / std::sqrt(1.0 + std::pow(slope, 2));
normal << -slope * normInv, 0.0, normInv;
```

core/src/utilities/geometry.cc
```diff
@@ -749,4 +749,41 @@ namespace jiminy
...
+                normal.segment<2>(0) = rot_mat * normal.segment<2>(0);
```
duburcqa: `.segment<2>(0)`is just `head<2>()`. And since `normal[1]` is 0.0, you could just do:
```cpp
normal.head<2>() = normal[0] * rot_mat.col(0);
```

For efficiency, I would suggest this overall formulation:
```cpp
normal << - slope * normInv * rot_mat.col(0), normInv;
```

Just add on before the original formula before simplifications. Otherwise it is hard to get from where it is coming at first glance.

core/src/utilities/geometry.cc
```diff
@@ -749,4 +749,41 @@ namespace jiminy
...
+        double stepWidth, double stepHeight, uint32_t stepNumber, double orientation)
+    {
+        const double interpDelta = 0.01;
+        const Eigen::Rotation2D<double> rot_mat(orientation);
+
+        return [stepWidth, stepHeight, stepNumber, rot_mat, interpDelta](
+                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
+        {
+            // Compute position in stairs reference frame
+            Eigen::Vector2d posRel = (rot_mat.inverse() * pos);
```
duburcqa: `Eigen::Vector2d posRel = rot_mat.inverse() * pos;`

duburcqa (commented):
Good job ! Almost there !

core/src/utilities/geometry.cc
```diff
@@ -749,4 +749,41 @@ namespace jiminy
...
+            {
+                stairIndex = 2 * stepNumber - stairIndex;
+                staircaseSlopeSign = -1;
+            }
+            height = stairIndex * stepHeight;
+            normal = Eigen::Vector3d::UnitZ();
+
+            // Avoid unsupported vertical edge
+            const double posRelOnStep = std::fmod(modPos, stepWidth) / stepWidth;
+            if (1 - posRelOnStep < interpDelta)
```
duburcqa: `1.0`

core/src/utilities/geometry.cc
mwulfman: Ok, I hesitated for this one. Also, is there a max slope for the terrain or can it just be as high as desired as long as it is not inf ? I've inspired myself from `tile2dInterp1d` for the implementation but the slope depends on the step height.

duburcqa: what do you mean ? it is exactly the same as you did yourself, just reformulated to avoid useless computations x)

duburcqa: There is a "max slope". The point is making sure that the normal is not pointing upward instantly after crossing the step. So it depends in theory of the integration tilmestep and max possible velocity. But a rough guess can be estimated.

duburcqa: Basically, there is a max slope (ie the "angle"). So the higher the step the longer the absolute interpolation length.

core/src/utilities/geometry.cc
mwulfman: What about this for documentation ? 
```
            if (1.0 - posRelOnStep < interpDelta)
            {
                const double slope = staircaseSlopeSign * stepHeight / interpDelta;
                // Update height
                height += slope * (posRelOnStep - (1.0 - interpDelta));

                // Compute the inverse of the normal's Euclidean norm
                const double normInv = 1.0 / std::sqrt(1.0 + std::pow(slope, 2));

                // Update normal vector
                // step 1. compute normal in stairs reference frame:
                // normal << -slope * normInv, 0.0, normInv;
                // step 2. Rotate normal vector in world reference frame:
                // normal.head<2>() = normal[0] * rot_mat.col(0);
                // Or simply in a single operation:
                normal << -slope * normInv * rot_mat.toRotationMatrix().col(0), normInv;
            }
```

duburcqa: Perfect. Just replace: `normal.head<2>() = normal[0] * rot_mat.col(0);` by `normal.head<2>() = rot_mat * normal.head<2>();`

core/src/utilities/geometry.cc
mwulfman: But in `tiles` or in the new `stairs` it is, the wider the tile/stairs, the longer the interpolation length.

core/src/utilities/geometry.cc
duburcqa: and `world reference frame:` by `world plane reference frame:`

core/src/utilities/geometry.cc
duburcqa: Yes for `tiles` it is, as it should be. But the **minimum slope** should be absolute not relative. A relative maximum slope of `0.01` was enforced for simplicity but it is indeed "wrong". In principle, the minimum interpolation delta should be computing dynamically according to the size of tiles / stairs.

duburcqa: feel free to fix this issue if you have time, it can indeed be consider a bug to some extent.

mwulfman: Ok, but if the "max slope" is not a fixed constant but depends on the integration timestep and max possible velocity, I understand that it is not trivial.

core/src/utilities/geometry.cc
mwulfman: Good catch, done.

core/src/utilities/geometry.cc
mwulfman: Patch applied.

core/src/utilities/geometry.cc
mwulfman: Fixed

core/src/utilities/geometry.cc
duburcqa: yes but a pessimistic fixed constant would be better a better guess than what we are currently doing I would say.

duburcqa: We can keep it like this for now if it is not blocking you.

mwulfman: It's not blocking me for now. I will reflect on it.

duburcqa:
CI is failing. You must replace `"playwright!=1.43"` by `"playwright<1.43"`  in `setup.py`.

mwulfman:
There's a Python linter issue now.

duburcqa:
Yes. No big deal. Merging.

PR #799, opened 2024-05-17
[core] Add Perlin ground profile generators.

mwulfman:
I am proposing to add a unidirection Perlin ground `HeightmapFunction`, 2D Perlin ground `HeightmapFunction` and a quantization methods to quantize any ground to certain z-values. 

- [x] Unidirectional Perlin Ground: 

periodicperlinground

- [x] 2D Perlin Ground
2D_random_perlin_ground

- [x] Pass the Perlin process by copy to the builder function
- [ ] Ground quantization method to quantized the ground to only certain level
- [ ] Unit tests
- [ ] Fix sample_state in gym-jiminy to account for the ground profile ?

### DIFF
--- a/core/include/jiminy/core/utilities/random.h
+++ b/core/include/jiminy/core/utilities/random.h
@@ -314,7 +314,7 @@ namespace jiminy
         /// \param[in] coeffs First row of the matrix to decompose.
         template<typename Derived>
         MatrixX<typename Derived::Scalar>
-        standardToeplitzCholeskyLower(const Eigen::MatrixBase<Derived> & coeffs);
+        standardToeplitzCholeskyLower(const Eigen::MatrixBase<Derived> & coeffs, double reg = 0.0);
     }
 
     class JIMINY_DLLAPI PeriodicGaussianProcess
@@ -327,7 +327,8 @@ namespace jiminy
 
         void reset(const uniform_random_bit_generator_ref<uint32_t> & g) noexcept;
 
-        double operator()(float t);
+        double operator()(double t);
+        double gradient(double t);
 
         double getWavelength() const noexcept;
         double getPeriod() const noexcept;
@@ -336,7 +337,7 @@ namespace jiminy
         const double wavelength_;
         const double period_;
 
-        const double dt_{0.02 * wavelength_};
+        const double dt_{0.1 * wavelength_};
         const Eigen::Index numTimes_{static_cast<int>(std::ceil(period_ / dt_))};
 
         /// \brief Cholesky decomposition (LLT) of the covariance matrix.
@@ -347,20 +348,42 @@ namespace jiminy
         ///          positive semi-definite Toepliz matrix, which means that the computational
         ///          complexity can be reduced even further using an specialized Cholesky
         ///          decomposition algorithm. See: https://math.stackexchange.com/q/22825/375496
-        Eigen::MatrixXd covSqrtRoot_{
-            internal::standardToeplitzCholeskyLower(Eigen::VectorXd::NullaryExpr(
+        ///          Ultimately, the algorithmic complexity can be reduced from O(n^3) to O(n^2),
+        ///          which is lower than the matrix multiplication itself.
+        ///
+        Eigen::MatrixXd covSqrtRoot_{internal::standardToeplitzCholeskyLower(
+            Eigen::VectorXd::NullaryExpr(
                 numTimes_,
                 [numTimes = static_cast<double>(numTimes_), wavelength = wavelength_](double i) {
                     return std::exp(-2.0 *
                                     std::pow(std::sin(M_PI / numTimes * i) / wavelength, 2));
-                }))};
+                }),
+            1e-9)};
+        Eigen::MatrixXd covJacobian_{Eigen::MatrixXd::NullaryExpr(
+            numTimes_,
+            numTimes_,
+            [numTimes = static_cast<double>(numTimes_),
+             wavelength = wavelength_,
+             period = period_](double i, double j)
+            {
+                return -2 * M_PI / period / std::pow(wavelength, 2) *
+                       std::sin(2 * M_PI / numTimes * (i - j)) *
+                       std::exp(-2.0 *
+                                std::pow(std::sin(M_PI / numTimes * (i - j)) / wavelength, 2));
+            })};
+
         Eigen::VectorXd values_{numTimes_};
+        Eigen::VectorXd grads_{numTimes_};
     };
 
     // **************************** Continuous 1D Fourier processes **************************** //
 
     /// \see Based on "Smooth random functions, random ODEs, and Gaussian processes":
-    ///      https://hal.inria.fr/hal-01944992/file/random_revision2.pdf */
+    ///      https://hal.inria.fr/hal-01944992/file/random_revision2.pdf
+    ///
+    /// \see For references about the derivatives of a Gaussian Process:
+    ///      http://herbsusmann.com/2020/07/06/gaussian-process-derivatives
+    ///      https://arxiv.org/abs/1810.12283
     class JIMINY_DLLAPI PeriodicFourierProcess
     {
     public:
@@ -371,7 +394,8 @@ namespace jiminy
 
         void reset(const uniform_random_bit_generator_ref<uint32_t> & g) noexcept;
 
-        double operator()(float t);
+        double operator()(double t);
+        double gradient(double t);
 
         double getWavelength() const noexcept;
         double getPeriod() const noexcept;
@@ -380,7 +404,7 @@ namespace jiminy
         const double wavelength_;
         const double period_;
 
-        const double dt_{0.02 * wavelength_};
+        const double dt_{0.1 * wavelength_};
         const Eigen::Index numTimes_{static_cast<Eigen::Index>(std::ceil(period_ / dt_))};
         const Eigen::Index numHarmonics_{
             static_cast<Eigen::Index>(std::ceil(period_ / wavelength_))};
@@ -389,13 +413,15 @@ namespace jiminy
             numTimes_,
             numHarmonics_,
             [numTimes = static_cast<double>(numTimes_)](double i, double j)
-            { return std::cos(2 * M_PI / numTimes * i * j); })};
+            { return std::cos(2 * M_PI / numTimes * i * (j + 1)); })};
         const Eigen::MatrixXd sinMat_{Eigen::MatrixXd::NullaryExpr(
             numTimes_,
             numHarmonics_,
             [numTimes = static_cast<double>(numTimes_)](double i, double j)
-            { return std::sin(2 * M_PI / numTimes * i * j); })};
+            { return std::sin(2 * M_PI / numTimes * i * (j + 1)); })};
+
         Eigen::VectorXd values_{numTimes_};
+        Eigen::VectorXd grads_{numTimes_};
     };
 
     // ***************************** Continuous 1D Perlin processes **************************** //
@@ -461,8 +487,8 @@ namespace jiminy
     private:
         const double period_;
 
-        std::vector<uint32_t> hashes_ =
-            std::vector<uint32_t>(static_cast<std::size_t>(period_ / wavelength_));
+        std::vector<float> grads_ =
+            std::vector<float>(static_cast<std::size_t>(period_ / wavelength_));
     };
 
     /// \brief  Sum of Perlin noise octaves.
@@ -496,7 +522,7 @@ namespace jiminy
     public:
         void reset(const uniform_random_bit_generator_ref<uint32_t> & g) noexcept;
 
-        double operator()(float t);
+        double operator()(double t);
 
         double getWavelength() const noexcept;
         std::size_t getNumOctaves() const noexcept;
--- a/core/include/jiminy/core/utilities/random.h
+++ b/core/include/jiminy/core/utilities/random.h
@@ -429,12 +429,17 @@ namespace jiminy
         virtual void reset(const uniform_random_bit_generator_ref<uint32_t> & g) noexcept;
 
         double operator()(const VectorN<double> & x) const;
-        // VectorN<double> grad(VectorN<double> const & x) const;
+        VectorN<double> grad(const VectorN<double> & x) const;
 
         double getWavelength() const noexcept;
 
     protected:
         virtual VectorN<double> grad(const VectorN<int32_t> & knot) const noexcept = 0;
+
+    private:
+        template<bool isGradient>
+        std::conditional_t<isGradient, VectorN<double>, double>
+        evaluate(const VectorN<double> & x) const;
 
     protected:
         const double wavelength_;
@@ -524,6 +529,7 @@ namespace jiminy
         void reset(const uniform_random_bit_generator_ref<uint32_t> & g) noexcept;
 
         double operator()(const VectorN<double> & x) const;
+        VectorN<double> grad(const VectorN<double> & x) const;
 
         double getWavelength() const noexcept;
         std::size_t getNumOctaves() const noexcept;
--- a/core/include/jiminy/core/utilities/random.h
+++ b/core/include/jiminy/core/utilities/random.h
@@ -314,7 +314,7 @@ namespace jiminy
         standardToeplitzCholeskyLower(const Eigen::MatrixBase<Derived> & coeffs, double reg = 0.0);
     }
 
-    class JIMINY_DLLAPI PeriodicTabularProcess
+    class JIMINY_TEMPLATE_DLLAPI PeriodicTabularProcess
     {
     public:
         explicit PeriodicTabularProcess(double wavelength, double period);
@@ -337,7 +337,7 @@ namespace jiminy
         Eigen::VectorXd grads_{numTimes_};
     };
 
-    class JIMINY_DLLAPI PeriodicGaussianProcess final : public PeriodicTabularProcess
+    class JIMINY_TEMPLATE_DLLAPI PeriodicGaussianProcess final : public PeriodicTabularProcess
     {
     public:
         explicit PeriodicGaussianProcess(double wavelength, double period);
@@ -385,7 +385,7 @@ namespace jiminy
     /// \see For references about the derivatives of a Gaussian Process:
     ///      http://herbsusmann.com/2020/07/06/gaussian-process-derivatives
     ///      https://arxiv.org/abs/1810.12283
-    class JIMINY_DLLAPI PeriodicFourierProcess final : public PeriodicTabularProcess
+    class JIMINY_TEMPLATE_DLLAPI PeriodicFourierProcess final : public PeriodicTabularProcess
     {
     public:
         explicit PeriodicFourierProcess(double wavelength, double period);
@@ -413,7 +413,7 @@ namespace jiminy
     uint32_t MurmurHash3(const void * key, int32_t len, uint32_t seed) noexcept;
 
     template<template<unsigned int> class DerivedPerlinNoiseOctave, unsigned int N>
-    class JIMINY_DLLAPI AbstractPerlinNoiseOctave
+    class JIMINY_TEMPLATE_DLLAPI AbstractPerlinNoiseOctave
     {
     public:
         template<typename Scalar>
@@ -441,7 +441,7 @@ namespace jiminy
     };
 
     template<unsigned int N>
-    class JIMINY_DLLAPI RandomPerlinNoiseOctave :
+    class JIMINY_TEMPLATE_DLLAPI RandomPerlinNoiseOctave :
     public AbstractPerlinNoiseOctave<RandomPerlinNoiseOctave, N>
     {
     public:
@@ -463,7 +463,7 @@ namespace jiminy
     };
 
     template<unsigned int N>
-    class JIMINY_DLLAPI PeriodicPerlinNoiseOctave :
+    class JIMINY_TEMPLATE_DLLAPI PeriodicPerlinNoiseOctave :
     public AbstractPerlinNoiseOctave<PeriodicPerlinNoiseOctave, N>
     {
     public:
@@ -518,7 +518,7 @@ namespace jiminy
     class AbstractPerlinProcess;
 
     template<template<unsigned int> class DerivedPerlinNoiseOctave, unsigned int N>
-    class JIMINY_DLLAPI AbstractPerlinProcess<DerivedPerlinNoiseOctave, N>
+    class JIMINY_TEMPLATE_DLLAPI AbstractPerlinProcess<DerivedPerlinNoiseOctave, N>
     {
     public:
         template<typename Scalar>
@@ -546,7 +546,7 @@ namespace jiminy
     };
 
     template<unsigned int N>
-    class JIMINY_DLLAPI RandomPerlinProcess :
+    class JIMINY_TEMPLATE_DLLAPI RandomPerlinProcess :
     public AbstractPerlinProcess<RandomPerlinNoiseOctave, N>
     {
     public:
--- a/core/include/jiminy/core/utilities/random.h
+++ b/core/include/jiminy/core/utilities/random.h
@@ -220,9 +220,9 @@ namespace jiminy
 
     template<typename Generator, typename Derived1, typename Derived2>
     std::enable_if_t<
-        (is_eigen_any_v<Derived1> ||
-         is_eigen_any_v<Derived2>)&&(!std::is_arithmetic_v<std::decay_t<Derived1>> ||
-                                     !std::is_arithmetic_v<std::decay_t<Derived2>>),
+        (is_eigen_any_v<Derived1> || is_eigen_any_v<Derived2>) &&
+            (!std::is_arithmetic_v<std::decay_t<Derived1>> ||
+             !std::is_arithmetic_v<std::decay_t<Derived2>>),
         Eigen::CwiseNullaryOp<
             scalar_random_op<float(const uniform_random_bit_generator_ref<uint32_t> &, float, float),
                              Generator &,
@@ -271,9 +271,9 @@ namespace jiminy
     /// optimizations enabled (level 01 is enough), probably due to inlining.
     template<typename Generator, typename Derived1, typename Derived2>
     std::enable_if_t<
-        (is_eigen_any_v<Derived1> ||
-         is_eigen_any_v<Derived2>)&&(!std::is_arithmetic_v<std::decay_t<Derived1>> ||
-                                     !std::is_arithmetic_v<std::decay_t<Derived2>>),
+        (is_eigen_any_v<Derived1> || is_eigen_any_v<Derived2>) &&
+            (!std::is_arithmetic_v<std::decay_t<Derived1>> ||
+             !std::is_arithmetic_v<std::decay_t<Derived2>>),
         Eigen::CwiseNullaryOp<
             scalar_random_op<float(const uniform_random_bit_generator_ref<uint32_t> &, float, float),
                              Generator &,
@@ -461,7 +461,8 @@ namespace jiminy
     private:
         const double period_;
 
-        std::array<uint8_t, 256> perm_{};
+        std::vector<uint32_t> hashes_ =
+            std::vector<uint32_t>(static_cast<std::size_t>(period_ / wavelength_));
     };
 
     /// \brief  Sum of Perlin noise octaves.
--- a/core/include/jiminy/core/utilities/random.h
+++ b/core/include/jiminy/core/utilities/random.h
@@ -331,7 +331,7 @@ namespace jiminy
         const double wavelength_;
         const double period_;
         const Eigen::Index numTimes_{static_cast<int>(std::ceil(period_ / (0.1 * wavelength_)))};
-        const double dt_{period_ / numTimes_};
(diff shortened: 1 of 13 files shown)


core/src/utilities/random.cc
```diff
@@ -735,4 +735,40 @@ namespace jiminy
...
+        {
+            // Compute the Perlin Process relative coordinate
+            Eigen::Vector2d posRel = (rot_mat.inverse() * pos).array() + offset;
+
+            // Compute the height
+            height = perlinProcess(std::abs(static_cast<float>(posRel[0])));
+
+            // Compute the numerical gradient of the Perlin Process
+            // and retrieve the normal
+            const double curvGrad = 0.5 * (
```
duburcqa: I'm not OK for merging a numerical approximation. A proper derivative operator should be added to `AbstractPerlinNoiseOctave`, based on the existing private method `grad`.

core/include/jiminy/core/utilities/random.h
```diff
@@ -536,6 +536,10 @@ namespace jiminy
                                           uint32_t sparsity,
                                           double orientation,
                                           uint32_t seed);
+
+    HeightmapFunction JIMINY_DLLAPI univariatePerlinGround(AbstractPerlinProcess & perlinProcess,
```
duburcqa: I'm in favour of instantiate the Perlin process directly inside the method, and pass the construct arguments to this method. It will reduce the set of processes being supported, but I'm not a huge fun of holding a reference to an object that may be modified later on by the user. Plus setting the seed of the Perlin process of an object that is not managed by the method itself is error-prone.

duburcqa:
> It would be nice to add a quantization options to quantized the ground to only certain level.

What do you mean ?

core/include/jiminy/core/utilities/random.h
mwulfman: So you would prefer something like this?

```
    HeightmapFunction unidirectionalPerlinGround(double wavelength,
                                                std::size_t numOctaves,
                                                 double orientation,
                                                 uint32_t seed)
    {

        const Eigen::Rotation2D<double> rot_mat(orientation);

       

        // Constant for numerical gradient
        const double h = 1e-6;
        const double offset = 100.0;

        return [wavelength, numOctaves, rot_mat, h, offset](
                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
        {
             // Set seed of Perlin Process
            PCG32 pcg32_generator=PCG32(seed);
            RandomPerlinProcess perlinProcess = RandomPerlinProcess(wavelength, numOctaves);
            perlinProcess.reset(pcg32_generator);

            // Compute the Perlin Process relative coordinate
            Eigen::Vector2d posRel = (rot_mat.inverse() * pos).array() + offset;

            // Compute the height
            height = perlinProcess(std::abs(static_cast<float>(posRel[0])));

            // Compute the numerical gradient of the Perlin Process
            // and retrieve the normal
            const double curvGrad = 0.5 * (
                perlinProcess(std::abs(static_cast<float>(posRel[0] + h)))
                - perlinProcess(std::abs(static_cast<float>(posRel[0] - h)))
                ) / h;
            normal << -curvGrad, 0.0, 1.0;
            normal.segment<2>(0) = rot_mat * normal.segment<2>(0);
            normal.normalize();
        };
    }
```
(If your forget about the gradient for now of course)
This would also lead to two functions, one for the `RandomPerlinProcess` and one for the `PeriodicPerlinProcess`.

core/src/utilities/random.cc
mwulfman: I agree, it's still in prototyping phase, I wanted to open the discussion with the PR.

mwulfman:
> > It would be nice to add a quantization options to quantized the ground to only certain level.
> 
> What do you mean ?

To be able to get this type of terrain:
Screenshot from 2024-05-17 18-31-57 (ref: https://fse.studenttheses.ub.rug.nl/30171/1/thesis.pdf)

core/include/jiminy/core/utilities/random.h
duburcqa: Yes. Just make the lambda function an actual function (only visible in `cpp` of course) and use `std::bind` to return a function storing its argument (same as `partial` in python). You could also return a `lambda` calling your method and nothing else, only to capture the arguments.

duburcqa:
I see. It looks like a quantisation based on the vertical value only. Something like `z = std::floor(z / delta) * delta`. The issue is that we still need to add small slopes...

I think `tiles` or `tile2dInterp1d` should be extended to support "tile-ization" of any ground profile. The advantage is that it does not require having access to the normal of the original ground profile !

duburcqa:
Are you planning to work on it ? If not, then I suggest closing this issue because I'm not planning to work on it either.

mwulfman:
Yes, I did a bit more work on it but didn’t push anything yet. I’m planning to work on/finish it in a couple of weeks.

duburcqa:
Ok perfect ! Let's keep it open then :)

duburcqa:
Apparently you made good progress ! Nice :)

What is your future plan ? You want to add quantisation ? Can you add a unit test checking if the gradient is right ?

core/src/utilities/random.cc
```diff
@@ -735,4 +775,88 @@ namespace jiminy
...
+        const Eigen::Rotation2D<double> rot_mat(orientation);
+
+        // Constant for avoiding discontinuity at (0.0, 0.0)
+        const double offset = 100.0;
+
+        return [wavelength, numOctaves, rot_mat, offset, seed](
+                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
+        {
+            // Set seed of Random Perlin Process
+            PCG32 pcg32_generator = PCG32(seed);
```
duburcqa: Move the instantiation of the Perlin process outside the lambda, and capture it (by copy) in the lambda.

core/src/utilities/random.cc
```diff
@@ -735,4 +775,88 @@ namespace jiminy
...
+
+            // Compute the Perlin Process relative coordinate
+            Eigen::Vector2d posRel = (rot_mat.inverse() * pos).array();
+            float delta = std::abs(static_cast<float>(posRel[0] + offset));
+
+            // Compute the height
+            height = perlinProcess(delta);
+
+            // Compute the gradient of the Perlin Process
+            // and retrieve the normal
```
duburcqa: Only one line for this comment should be enough.

core/src/utilities/random.cc
```diff
@@ -735,4 +775,88 @@ namespace jiminy
...
+        const Eigen::Rotation2D<double> rot_mat(orientation);
+
+        // Constant for avoiding discontinuity at (0.0, 0.0)
+        const double offset = 100.0;
+
+        return [wavelength, period, numOctaves, rot_mat, offset, seed](
+                   const Eigen::Vector2d & pos, double & height, Eigen::Vector3d & normal) -> void
+        {
+            // Set seed of Periodic Perlin Process
+            PCG32 pcg32_generator = PCG32(seed);
```
duburcqa: same

duburcqa:
> leads to floating point exception (core dumped)

The problem is not that it asserts to false because asserts are disabled for release build type. The issue is rather the contrary, we should systematically check this condition and raise an exception if not satisfying this condition causes segfault.

I will issue a fix for this.

mwulfman:
> Apparently you made good progress ! Nice :)
> 
> What is your future plan ? You want to add quantisation ? Can you add a unit test checking if the gradient is right ?

Yes, I still need to do :
 - some update of the code (following your comments as well),
 - add quantisation on any ground,
 - since I'm already into Perlin noise, I'm thinking of adding 2d perlin noise/ground
 - Yes, add unit test for the gradient. I was thinking of comparing the gradient with the numerical gradient, were you thinking of any other tests?
 - Maybe update `sample_state` in gym-jiminy to account for the HeightMap.

duburcqa:
> I was thinking of comparing the gradient with the numerical gradient, were you thinking of any other tests?

Nope it is fine :)

> Maybe update sample_state in gym-jiminy to account for the HeightMap.

I'm afraid doing this is not straightforward. It has been a long-lasting issue as you know. Once everything else is working, I could help you on this part to improve the current heuristic that has not sufficient in many cases.

core/src/utilities/random.cc
mwulfman: I am having trouble with capturing the Perlin process by copy since in the code, the copy is explicitly disabled:
```       
JIMINY_DISABLE_COPY(AbstractPerlinProcess)
```

duburcqa: Try using move semantic then ! I guess it should work: `[generator=std::move(pcg32_generator)] (...) {...}`. Otherwise, I can implement the copy constructor, it should be fairly straightforward. It was just disable by simplicity.

mwulfman: Ok thanks, I wasn't sure if I could remove this part or if it was on purpose. I have implemented the copy constructor if it works for you.

core/src/utilities/random.cc
```diff
@@ -382,6 +382,94 @@ namespace jiminy
...
+        const double dRatio = fade_derivative(dtLeft);
+
+        /* Compute gradients at knots, and perform linear interpolation between them to get value
+           at current phase.*/
+        const double yLeft = grad(phaseIndexLeft, dtLeft);
+        const double yRight = grad(phaseIndexRight, dtRight);
+        return (yRight - yLeft) * dRatio + lerp(ratio, yLeft / dtLeft, yRight / dtRight);
+    }
+
+    double AbstractPerlinNoiseOctave::operator()(double x, double y) const
```
duburcqa: I don't understand why do you need all. 2D Perlin process is not the same as the sum of 1D processes along x-axis and y-axis respectively ? If so, just do this.

core/src/utilities/random.cc
```diff
@@ -746,4 +950,136 @@ namespace jiminy
             }
         };
     }
-}
\ No newline at end of file
+
+    template<typename PerlinProcessType>
+    void computePerlinHeightmap(PerlinProcessType perlinProcess,
```
duburcqa: Just make this signature more general to support specifying pairs `(axis, PerlinProcess)`, this way:
```
    template<typename PerlinProcessType, int N>
    void computePerlinHeightmap(std::array<std::pair<Eigen::Vector3, PerlinProcessType>, N> perlinProcesses,
                                ...
```

duburcqa:
To many code duplication. There should be no need to implement anything specific for the 2D scenario.

core/src/utilities/random.cc
duburcqa: Then all you have to do is to project the current pose only each axes and aggregate the result. You can do this automatically using `sumHeightmaps`, which has been designed for this exact purpose.

duburcqa:
I'm going to help you on this, so that we can close this long-lasting PR.

core/src/utilities/random.cc
mwulfman: random_perlin_ground_bidirectional

2D Perlin process is not the same as the sum of 1D processes along x-axis and y-axis. Here's a photo of the sum (to compare with the photo in the description). That's what I did first and then realized quickly it wasn't what I was looking for.

core/src/utilities/random.cc
mwulfman: Like shown in the photo above, I don't think that's what we want.

mwulfman:
> I'm going to help you on this, so that we can close this long-lasting PR.

If you have time for it, I'm pretty happy to get your help. I didn't have much time to work on it the last few weeks, but hoping to finish it this week.

core/src/utilities/random.cc
duburcqa: Ok now I see the issue. It is indeed more tricky than expected at first glance. I'm going to think about it to see if there is a way to get the same result without duplicating everything. The big issue is that the contribution of each process independent on each axis. 

I'm wondering if passing the 2D position encoding (hash) to two independent random processes should do the trick. I will try that.

core/src/utilities/random.cc
```diff
@@ -746,4 +950,136 @@ namespace jiminy
...
+    {
+        const Eigen::Rotation2D<double> rot_mat(orientation);
+
+        // Set seed of Random Perlin Process
+        PCG32 pcg32_generator = PCG32(seed);
+        RandomPerlinProcess perlinProcess = RandomPerlinProcess(wavelength, numOctaves);
+        perlinProcess.reset(pcg32_generator);
+
+        // Constant for avoiding discontinuity at (0.0, 0.0)
+        const double offset = 100.0;
```
duburcqa: What is this ?

core/src/utilities/random.cc
```diff
@@ -485,6 +608,34 @@ namespace jiminy
         return 2.0 * grad * delta;
     }
 
+    double PeriodicPerlinNoiseOctave::grad2D(
+        int32_t ix, int32_t iy, double x, double y) const noexcept
+    {
+        // Compute random gradient
+        // Get knot
+        uint64_t knot = (static_cast<uint64_t>(ix) << 32) | static_cast<uint32_t>(iy);
```
duburcqa: I'm not convinced by this. The periodicity is supposed to have the spatial interpretation in physical world. A repeating pattern is supposed to be clearly visible. I'm wondering what it looks like in this case. In the case of periodic Perlin noise, I have the feeling that `sumHeightmaps` is the way to go. Anyway, I don't think it really makes sense for ground profile generation.

core/src/utilities/random.cc
mwulfman: It's a non-ideal fix. Since the 1d perlin process is defined only on t >0, I needed a way to have it well defined for t<0 as well. Using the absolute value of t doesn't work since you end up with a discontinuity at t=0. So instead I shifted the perlin process by a offset. But it doesn't change the fact that the perlin process remains undefined for t < offset.

core/src/utilities/random.cc
duburcqa: After check, it is indeed more tricky than that and you really need a 2D implementation. Position encoding is not enough per se. I forgot about random Perlin processes were evaluated in the first place. Since it is an interpolation, it is always necessary to have a dedicated 2D implementation.

core/src/utilities/random.cc
duburcqa: Ok I will check if it can be refactored to work with negative values.

core/src/utilities/random.cc
duburcqa: One option would be to do:
```cpp
uint32_t deltaPeriod = static_cast<uint32_t>(period_ / wavelength_)
ix %= deltaPeriod;
iy %= deltaPeriod;
uint64_t knot = (static_cast<uint64_t>(ix) << 32) | static_cast<uint32_t>(iy);
const double s = perm_[knot] / 256.0;
```

By the way, I think there was an issue in the 1D case if `deltaPeriod > 256`

duburcqa:
Ok I skimmed through the code. Now I remember. Since those processes at not used for anything inside core itself, I'm going to refactor them a bit. In particular to replace `operator()` by `evaluate` and add `evaluate_gradient` which will return both the value and its gradient. Some internal methods also need a better name, such as `grad` that is confusing now. Then, I will rather template them with the dimensionality of the process, with only 1 and 2 being supported at the moment. I hope this will make everything more simple and clear, while avoid redundancy.

mwulfman:
Since you're taking over for the PerlinNoise part, I started creating some unit tests (using the current methods' name) and fixed a typo in the 1D gradient computation. There is also a bug in the 2D gradient computation.

duburcqa:
Ok nice, so far I have been working on adding the gradient feature on all the other random processes so as to provide a unified API for all of them, without slowing down computations. I have also fixed stuff like negative time, and fixed bugs on non-perlin processes.

duburcqa:
I will commit something this morning.

core/src/utilities/random.cc
```diff
@@ -862,7 +871,7 @@ namespace jiminy
                             double orientation,
                             uint32_t seed)
     {
-        if ((0.01 < interpDelta.array()).any() || (interpDelta.array() > size.array() / 2.0).any())
+        if ((0.01 > interpDelta.array()).any() || (interpDelta.array() < size.array() / 2.0).any())
```
duburcqa: still wrong.

duburcqa:
OK I'm going to rebase your branch on mine and push the whole stack on this PR.

core/src/utilities/random.cc
mwulfman: Yes indeed, I've done it too quickly x)

duburcqa:
For the record, it is taking a bit more time than expected because I'm going for the N-dimensional implementation.

duburcqa:
Just for you to now, I'm still on it. It is almost done. It was a bit more tricky than expected, but now it works in N-d without gradient.

duburcqa:
I'm done with the gradient. Adding terrain generation.

duburcqa:
Ready to merge. Missing stuff if any could be added later.

duburcqa (dismissed):

mwulfman:
Can you try creating a periodic_perlin_ground on your side ? I am having a bit of trouble doing so:
```
In [15]: periodic_perlin_ground?
Docstring: periodic_perlin_ground( (float)wavelength, (float)period, (int)num_octaves, (int)seed) -> HeightmapFunction
Type:      function

In [16]: periodic_perlin_ground(1.0,2.0,6,0)
---------------------------------------------------------------------------
LogicError                                Traceback (most recent call last)
Cell In[16], line 1
----> 1 periodic_perlin_ground(1.0,2.0,6,0)

LogicError: jiminy::PeriodicPerlinNoiseOctave<2>::PeriodicPerlinNoiseOctave(/Users/mathiasw/Code/jiminy/core/include/jiminy/core/utilities/random.hxx:457):
'period' must be larger than 'wavelength'.
```

python/jiminy_pywrap/src/generators.cc
```diff
@@ -93,17 +92,95 @@ namespace jiminy::python
...
+                    (bp::arg("self"), "generator")))
+                .ADD_PROPERTY_GET("wavelength", &DerivedPerlinProcess::getWavelength)
+                .ADD_PROPERTY_GET("num_octaves", &DerivedPerlinProcess::getNumOctaves);
+            // clang-format on
+        }
+
+        static void expose()
+        {
+            bp::class_<RandomPerlinProcess<N>,
+                       // bp::bases<AbstractPerlinProcess<RandomPerlinNoiseOctave, N>>,
```
mwulfman, line 156: Do you need to keep this commented line ?

python/jiminy_pywrap/src/generators.cc
```diff
@@ -93,17 +92,95 @@ namespace jiminy::python
...
+                       std::shared_ptr<RandomPerlinProcess<N>>,
+                       boost::noncopyable>(
+                toString("RandomPerlinProcess", N, "D").c_str(),
+                bp::init<double, uint32_t>(
+                    (bp::arg("self"), "wavelength", bp::arg("num_octaves") = 6U)))
+                .def(PyPerlinProcessVisitor<N>());
+
+            bp::class_<PeriodicPerlinProcess<N>,
+                       // bp::bases<AbstractPerlinProcess<PeriodicPerlinNoiseOctave, N>>,
```
mwulfman, line 165: Same

core/unit/random_test.cc
```diff
@@ -0,0 +1,95 @@
...
+
+    Eigen::Vector2d pos{5.43, 7.12};
+    Eigen::Vector2d finiteDiffGrad{(octave(pos + DELTA * Eigen::Vector2d::UnitX()) -
+                                    octave(pos - DELTA * Eigen::Vector2d::UnitX())) /
+                                       (2 * DELTA),
+                                   (octave(pos + DELTA * Eigen::Vector2d::UnitY()) -
+                                    octave(pos - DELTA * Eigen::Vector2d::UnitY())) /
+                                       (2 * DELTA)};
+    ASSERT_TRUE(finiteDiffGrad.isApprox(octave.grad(pos), TOL));
+}
```
mwulfman, line 95: Shouldn't we add test on the PerlinProcess as well ?

mwulfman:
Also, I had started working on sample_state and added:

gym_jiminy/common/envs/locomotion.py, line 239..
```

        # Compute the height of the freeflyer in neutral configuration
        # TODO: Take into account the ground profile.
        q_init, _ = self._sample_state()

        ground_profile = self.simulator.get_options()["world"]["groundProfile"]
        self._height_neutral = q_init[2] - ground_profile(q_init[:2])[0]
 ```
 
gym_jiminy/common/envs/generic.py, line 1398
 ```
 
         # Get ground profile
        ground_profile = self.simulator.get_options()["world"]["groundProfile"]

        # Lowest contact point from ground
        contact_frames_xy_positions = np.stack([self.robot.pinocchio_model.frames[frame_index].placement.translation[:2]
                                    for frame_index in self.robot.contact_frame_indices])
        contact_frames_z_positions = np.stack([self.robot.pinocchio_model.frames[frame_index].placement.translation[2]
                                    for frame_index in self.robot.contact_frame_indices])
        ground_height_contact_frames = np.zeros((len(self.robot.contact_frame_indices),))
        query_heightmap(ground_profile, contact_frames_xy_positions, ground_height_contact_frames)
        delta_height_from_ground = contact_frames_z_positions - ground_height_contact_frames

        if (delta_height_from_ground >= 0.0).all():
            # Lowest=Closest contact points needs to be into contact
            # with the ground:
            q[2] -= min(delta_height_from_ground)
        else:
            # From the points below ground, deepest one needs to be in contact
            # with the ground
            delta_height = delta_height_from_ground[delta_height_from_ground < 0.0]
            q[2] = q[2] - delta_height.min()
```
Either there is a bug in this code or there is something that I've been missing because it's not working as expected.

python/jiminy_pywrap/src/generators.cc
duburcqa: Yes, as a reminder that it derives from `AbstractPerlinProcess<...>` but inheritance is not declared official as the base class itself is not exposed to the end-user.

core/unit/random_test.cc
duburcqa: Ideally we should, but I got tired on the way x)

duburcqa:
> Either there is a bug in this code or there is something that I've been missing because it's not working as expected.

No need to make two branches based on the sign of `delta_height_from_ground`. This should be fine:
```python
        # Get ground profile
       ground_profile = self.simulator.get_options()["world"]["groundProfile"]

       # Lowest contact point from ground
       contact_frames_xy_positions = np.stack([
           self.robot.pinocchio_model.frames[frame_index].placement.translation[:2]
           for frame_index in self.robot.contact_frame_indices])
       contact_frames_z_positions = np.stack([
           self.robot.pinocchio_model.frames[frame_index].placement.translation[2]
           for frame_index in self.robot.contact_frame_indices])
       ground_height_contact_frames = np.zeros((len(self.robot.contact_frame_indices),))
       query_heightmap(ground_profile, contact_frames_xy_positions, ground_height_contact_frames)
       delta_height_from_ground = contact_frames_z_positions - ground_height_contact_frames

       # Make sure that there is exactly one point in contact with the ground, 
       # all the others being above the ground
       q[2] -= min(delta_height_from_ground)
```

Apart from that, it looks good. What is the issue ?

mwulfman:
Screenshot 2024-07-12 at 10 26 20 PM

The issue is that it is flying when I test it on Perlin ground...

Also the API changed on query_heighmap so it needs to be:

```
        contact_frames_xy_positions = np.stack([self.robot.pinocchio_model.frames[frame_index].placement.translation[:2]
                                    for frame_index in self.robot.contact_frame_indices], -1)
 ```

**Comment by duburcqa** [comment]
> Can you try creating a periodic_perlin_ground on your side ? I am having a bit of trouble doing so:

It is expected, but the error message is misleading. I will fix it. In fact, each octave has a larger wavelength than the "previous" one, so if you add too many octaves, at some point it can exceed the period.

**Comment by mwulfman** [comment]
> > Can you try creating a periodic_perlin_ground on your side ? I am having a bit of trouble doing so:
> 
> It is expected, but the error message is misleading. I will fix it. In fact, each octave has a larger wavelength than the "previous" one, so if you add too many octaves, at some point it can exceed the period.

Is it standard implementation or some implementation use the other convention (each octave has a higher frequency) ?

**Comment by duburcqa** [comment]
> The issue is that it is flying when I test it on Perlin ground...

Did you updated the forward kinematic by calling `update_quantities` or `framesForwardKinematics` ?
```python
# Set robot in neutral configuration
q = self._neutral()
pin.framesForwardKinematics(self.robot.pinocchio_model, self.robot.pinocchio_data, q)
```

duburcqa:
> Is it standard implementation or some implementation use the other convention (each octave has a higher frequency) ?

It is standard, look for Perlin Noise "persistence" and "lacunarity".

mwulfman:
> > The issue is that it is flying when I test it on Perlin ground...
> 
> Did you updated the forward kinematic by calling `update_quantities` or `framesForwardKinematics` ?
> 
> ```python
> # Set robot in neutral configuration
> q = self._neutral()
> pin.framesForwardKinematics(self.robot.pinocchio_model, self.robot.pinocchio_data, q)
> ```

It shouldn't be needed since it is already called after `sample_state` in `reset`:
```
        # Sample the initial state and reset the low-level engine
        q_init, v_init = self._sample_state()
        if not jiminy.is_position_valid(self.robot.pinocchio_model, q_init):
            raise RuntimeError(
                "The initial state provided by `_sample_state` is "
                "inconsistent with the dimension or types of joints of the "
                "model.")

        # Set robot in initial configuration
        pin.framesForwardKinematics(
            self.robot.pinocchio_model, self.robot.pinocchio_data, q_init)
```

duburcqa:
> Also the API changed on query_heighmap so it needs to be:

Yes, it changes for leveraging vectorisation more efficiently and hence improve performance.

duburcqa:
> It shouldn't be needed since it is already called after sample_state in reset:

Indeed. If you called refresh afterward then there is something very strange happening.

mwulfman:
Here is the code snippet I use. I simply call `env.render()` at breakpoint :
```python
import numpy as np

from gym_jiminy.envs.anymal import ANYmalPDControlJiminyEnv
from jiminy_py import tree
from jiminy_py.core import periodic_perlin_ground, random_perlin_ground


if __name__ == '__main__':
    env = ANYmalPDControlJiminyEnv()
    engine_options = env.simulator.get_options()
    orientation = 0.0
    wavelength= 1.0
    num_octaves = 8
    seed = 42
    engine_options["world"]["groundProfile"] =  random_perlin_ground(wavelength, num_octaves, seed)
    env.simulator.set_options(engine_options)
    for i in range(5):
        obs, _ = env.reset(seed = i)
        terminated, truncated = False, False
        while not (terminated or truncated):
            assert not any(np.isnan(e).any() for e in tree.flatten(obs)), "Nan in observation"

            obs, _, terminated, truncated, _ = env.step(0.0*env.action_space.sample())
            breakpoint()
            if terminated or truncated:
                print(f"terminated: {terminated}, truncated: {truncated}")
        env.replay(backed="panda3d")
        env.stop()

```

duburcqa:
Here is your issue:
```python
            contact_positions = np.stack([
                self.robot.pinocchio_data.oMf[frame_index].translation
                for frame_index in self.robot.contact_frame_indices], -1)
```

`self.robot.pinocchio_model.frames[frame_index].placement` is the relative pose of the frame wrt to its parent (it is constant).

duburcqa:
I can confirm it is working with:

```python
            engine_options = self.simulator.get_options()
            ground_fun = engine_options['world']['groundProfile']

            # Lowest contact point from ground
            contact_positions = np.stack([
                self.robot.pinocchio_data.oMf[frame_index].translation
                for frame_index in self.robot.contact_frame_indices], -1)
            contact_heights = np.zeros((contact_positions.shape[-1],))
            jiminy.query_heightmap(
                ground_fun, contact_positions[:2], contact_heights)
            delta_height_from_ground = contact_positions[2] - contact_heights

            # Make sure that there is exactly one point in contact with the
            # ground, all the others being above the ground.
            q[2] -= delta_height_from_ground.min()
```

mwulfman:
Oh yes, that's a nice catch. It is still floating in the air on my end..

duburcqa:
It is fine on my side ! (I mean with "your" patch to make the robot touch the ground, it is not merged for now as it is not generic enough, it should also consider collision bodies, even if they are not properly supported at the time being that are used in ant environment for instance). Ideally, `compute_freeflyer_state_from_fixed_body` should be fixed to what it is supposed to do, but doing it properly it is quite difficult.

image

duburcqa:
I need to profile this code, because I'm not happy at all with the performance:
```python
import numpy as np
from jiminy_py.core import random_perlin_ground, query_heightmap

ground_fun = random_perlin_ground(wavelength=1.0, num_octaves=8, seed=42)
positions = np.random.rand(3, 100)

%timeit query_heightmap(ground_fun, positions[:2], positions[2])
76.1 µs ± 680 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

I consider this as extremely bad. I would expect something like 5us at max. There is a lot of redundant computations everywhere: the gradient is computed independently from the height, and the gradient at knots is never cached even if the query point is only shifted by 1mm.

duburcqa:
image

Damn... most of the time is spend it computing sincos and pow... sounds tricky to optimise...

PR #819, opened 2024-06-27
 [jiminy_py] Fix 'tree.unflatten_as' mixing up key order for 'gym.spaces.Dict'.

### LINKED ISSUE #818: [gym][wrappers] StackObservation is not preserving the order for OrderedDict observation spaces.
observation_stack.StackObservation changes the key order of an OrderedDict observation_space while preserving the correct key order of observation. If it is used in conjunction with FlattenObservation, we find ourselves with env.observation and env.observation_spaces arrays with indices that doesn't correspond to each other. A side effect is then that the obs (after step) becomes different from env.observation because it is then clipped wrongly by a faulty env.observation_space. 

Reproduction script:
```
from gym_jiminy.common.wrappers import FlattenObservation, StackObservation
from gym_jiminy.envs.atlas import AtlasPDControlJiminyEnv

env = StackObservation(AtlasPDControlJiminyEnv(),skip_frames_ratio=-1,
            num_stack=2,
            nested_filter_keys=[["t"]],
        )
_, _ = env.reset(seed = 0)

print("======= env.observation keys() ==========")
for key in env.observation:
    print(key)
print("======= env.observation_space keys() ==========")
for key in env.observation_space:
    print(key)


env = FlattenObservation(StackObservation(AtlasPDControlJiminyEnv(),skip_frames_ratio=-1,
            num_stack=2,
            nested_filter_keys=[["t"]],
        ))
_, _ = env.reset(seed = 0)

N = 100
for _ in range(N):
    env.step(env.action_space.sample())
obs, _, _, _, _ = env.step(env.action_space.sample())

print("=========== Bounds of unflatten (stacked) env.env.observation space of 't' ========" )
print(env.env.observation_space["t"].low)
print(env.env.observation_space["t"].high)

print("=========== Bounds of flatten env.observation_space first 2 elements ========" )
print(env.observation_space.low[:2])
print(env.observation_space.high[:2])

print("=========== env.observation's first 2 elements ========" )
print(env.observation[:2])
print("=========== obs' 2 first elements ========" )
print(obs[:2])
print("======= The above two lines should be equals ! ===========")
```

Here is a fix if you think this solution works for you:
```
def _unflatten_as(data: StructNested[Any],
                  data_leaf_it: Iterator) -> StructNested[Any]:
    """Internal method un-flattening a sequence into a nested data structure by
    calling itself recursively on each top-level nodes that are not leaves and
    fetching the next leaf value otherwise.
    """
    data_type = type(data)
    if issubclass_mapping(data_type):  # type: ignore[arg-type]
        return data_type(OrderedDict({  # type: ignore[call-arg]
            key: _unflatten_as(value, data_leaf_it)
            for key, value in data.items()  # type: ignore[union-attr]
        }))
    if issubclass_sequence(data_type):  # type: ignore[arg-type]
        return data_type(tuple(  # type: ignore[call-arg]
            _unflatten_as(value, data_leaf_it) for value in data
        ))
    return next(data_leaf_it)
```

The solution just wrapped the created dict by the OrderedDict class.

mwulfman:
Resolves #818

### DIFF
--- a/python/jiminy_py/src/jiminy_py/tree.py
+++ b/python/jiminy_py/src/jiminy_py/tree.py
@@ -176,8 +176,9 @@ def _unflatten_as(data: StructNested[Any],
         flat_items = [(key, _unflatten_as(value, data_leaf_it))
             for key, value in data.items()]  # type: ignore[union-attr]
         try:
-            # Initialisation from dict cannot be the default path as `gym.spaces.Dict`
-            # would sort keys in this specific scenario, which must be avoided.
+            # Initialisation from dict cannot be the default path as
+            # `gym.spaces.Dict` would sort keys in this specific scenario,
+            # which must be avoided.
             return data_type(flat_items)  # type: ignore[call-arg]
         except (ValueError, RuntimeError):
             # Fallback to initialisation from dict in the rare event of
--- a/python/jiminy_py/src/jiminy_py/tree.py
+++ b/python/jiminy_py/src/jiminy_py/tree.py
@@ -173,19 +173,15 @@ def _unflatten_as(data: StructNested[Any],
     """
     data_type = type(data)
     if issubclass_mapping(data_type):  # type: ignore[arg-type]
+        flat_items = [(key, _unflatten_as(value, data_leaf_it))
+            for key, value in data.items()]  # type: ignore[union-attr]
         try:
-            return data_type([  # type: ignore[call-arg]
-                (key, _unflatten_as(value, data_leaf_it))
-                for key, value in data.items()  # type: ignore[union-attr]
-            ])
+            return data_type(flat_items)  # type: ignore[call-arg]
         except (ValueError, RuntimeError):
             # Fallback to initialisation from dict in the rare event of
             # a container type not supporting initialisation from a
             # sequence of key-value pairs.
-            return data_type({  # type: ignore[call-arg]
-                key: _unflatten_as(value, data_leaf_it)
-                for key, value in data.items()  # type: ignore[union-attr]
-            })
+            return data_type(dict(flat_items))  # type: ignore[call-arg]
     if issubclass_sequence(data_type):  # type: ignore[arg-type]
         return data_type(tuple(  # type: ignore[call-arg]
             _unflatten_as(value, data_leaf_it) for value in data
--- a/python/jiminy_py/src/jiminy_py/tree.py
+++ b/python/jiminy_py/src/jiminy_py/tree.py
@@ -19,6 +19,7 @@ of use and versatility rather with optimal performance.
 from functools import lru_cache
 from itertools import chain, starmap
 from collections.abc import (Mapping, ValuesView, Sequence, Set)
+from collections import OrderedDict
 from typing import (
     Any, Union, Mapping as MappingT, Iterable, Iterator as Iterator, Tuple,
     TypeVar, Callable, Type)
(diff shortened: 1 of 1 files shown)


python/jiminy_py/src/jiminy_py/tree.py
```diff
@@ -173,10 +174,10 @@ def _unflatten_as(data: StructNested[Any],
     """
     data_type = type(data)
     if issubclass_mapping(data_type):  # type: ignore[arg-type]
-        return data_type({  # type: ignore[call-arg]
+        return data_type(OrderedDict({  # type: ignore[call-arg]
```
duburcqa: I don't get it. Why `data` is not already an `OrderedDict` instance ? It feels wrong to first create a dict, then create an ordered dict and finally create a `data_type` instance...

mwulfman: When you work with gymnasium environment you can get a gymnasium.spaces.Dict.

duburcqa: Ok so the issue is actually `gymnasium.spaces.Dict` sorting keys when initialised from `dict` instead of `OrderedDict`, which fells wrong in the first place.

mwulfman: Yes, so I'm not sure how best to handle this case.

duburcqa: while `gymnasium<1.0` is not enforcing key sorting, `gymnasium>=1.0` does, which means that this trick is going to break at some point...

duburcqa: From what I see, it should work as expected when initialised from a sequence of tuple, so I would rather suggest something like this:
```python
try:
    # Initialisation from dict cannot be the default path as `gym.spaces.Dict`
    # would sort keys in this specific scenario, which must be avoided.
    data_type([(key, _unflatten_as(value, data_leaf_it)) for key, value in data.items()])
except (ValueError, RuntimeError):
    # Fallback to initialisation from dict in the rare event of a container 
    # type not supporting initialisation from a sequence of key-value pairs.
    data_type({key: _unflatten_as(value, data_leaf_it) for key, value in data.items()})
```

It should be both robust and fairly efficient since most if at all containers support the main pathway.

duburcqa: I would be worth checking if the fallback is used at the point when running the unit tests. I guess not, but still worth to add it for better reliability.

mwulfman: I've applied your suggestion and ran the unit tests. The fallback is not being used during the tests. And from what I see (the reproduction script), it works as expected.

python/jiminy_py/src/jiminy_py/tree.py
```diff
@@ -173,10 +173,19 @@ def _unflatten_as(data: StructNested[Any],
     """
     data_type = type(data)
     if issubclass_mapping(data_type):  # type: ignore[arg-type]
-        return data_type({  # type: ignore[call-arg]
-            key: _unflatten_as(value, data_leaf_it)
-            for key, value in data.items()  # type: ignore[union-attr]
-        })
+        try:
```
duburcqa, line 180: I would rather suggest something like this to avoid redundant computations in fallback:
```
flat_items = [(key, _unflatten_as(value, data_leaf_it)) for key, value in data.items()]
try:
    # Initialisation from dict cannot be the default path as `gym.spaces.Dict`
    # would sort keys in this specific scenario, which must be avoided.
    data_type(flat_items)
except (ValueError, RuntimeError):
    # Fallback to initialisation from dict in the rare event of a container 
    # type not supporting initialisation from a sequence of key-value pairs.
    data_type(dict(flat_items))
```

python/jiminy_py/src/jiminy_py/tree.py
duburcqa: Perfect, I made one extra minor suggestion and we are good.

python/jiminy_py/src/jiminy_py/tree.py
mwulfman: Yes it's cleaner

PR #891, opened 2025-11-02
[core] Issue with robot/model, you can't remove the added frames with removeFrames

### LINKED ISSUE #890: [core] Issue with  robot/model, you can't remove the added frames with removeFrames
It appears the safeguard part of the code in Model::removeFrames do the opposite of what it is supposed to do.

If not specifying the frame type, the frames you add using addFrame have the default type `OP_FRAME`. But then, when you tried removing those frames, without specifying `filter` argument, it takes by default `OP_FRAME`, meaning you can't remove the `OP_FRAME`.
The python api only exposes the add_frame and remove_frames without being able to specify the `type` or `filter` argument for the respective function. As a result, the added framed can't be removed.

This looks like a typo as we should be able to remove the `OP_FRAME`.

mwulfman:
Resolve #890

### DIFF
--- a/core/src/robot/model.cc
+++ b/core/src/robot/model.cc
@@ -479,7 +479,7 @@ namespace jiminy
     }
 
     void Model::removeFrames(const std::vector<std::string> & frameNames,
-                             const std::vector<pinocchio::FrameType> & filter)
+                             const std::vector<pinocchio::FrameType> & allowed)
     {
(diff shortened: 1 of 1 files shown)


duburcqa:
Thank you for this fix. However I'm afraid the CI is probably broken since it was not used for quite a while, so it will be necessary to fix it before being able to merge this PR :/

